PR #9, opened 2021-06-12
Changed instructions on creating credential files

jamesarnall:
The original instructions specified putting the gspread credentials file in the default location, but the code looks for a file in another place. The README now reflects the code's behavior.


loganwilliams:
Looks great, thank you for the contribution!

PR #12, opened 2022-02-13
add ability to archive twitter images from tweet url

msramalho:
this needs to be tested before, perhaps we can setup a test sheet with an example of each type of archivable link, then we can trigger the archiver locally and make sure all works as expected

### DIFF
--- a/auto_archive.py
+++ b/auto_archive.py
@@ -1,6 +1,5 @@
 import gspread
 import youtube_dl
-from pathlib import Path
 import sys
 import datetime
 import boto3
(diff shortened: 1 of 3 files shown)


auto_archive.py
```diff
@@ -117,9 +117,29 @@ def get_thumbnails(filename, s3_client, duration = None):
 
     return (key_thumb, thumb_index_cdn_url)
 
+def fetch_and_upload_to_s3(url, key, s3_client, check_if_exists=False, headers={}, status="success"):
```
msramalho, line 120: isolated this code into a function since it is re-used

PR #18, opened 2022-03-09
configurable column names + minor improvements

msramalho:


archivers/telegram_archiver.py
```diff
@@ -43,8 +43,10 @@ def download(self, url, check_if_exists=False):
                 images += urls
 
             page_cdn, page_hash, thumbnail = self.generate_media_page(images, url, html.escape(str(t.content)))
+            time_elements = s.find_all('time')
+            timestamp = time_elements[0].get('datetime') if len(time_elements) else None
 
-            return ArchiveResult(status="success", cdn_url=page_cdn, screenshot=screenshot, hash=page_hash, thumbnail=thumbnail, timestamp=s.find_all('time')[0].get('datetime'))
+            return ArchiveResult(status="success", cdn_url=page_cdn, screenshot=screenshot, hash=page_hash, thumbnail=thumbnail, timestamp=timestamp)
```
msramalho, line 49: was throwing index out of range

auto_archive.py
```diff
@@ -104,62 +105,59 @@ def process_sheet(sheet, header=1):
             archivers.WaybackArchiver(s3_client, driver)
         ]
 
-        values = gw.get_values()
         # loop through rows in worksheet
         for row in range(1 + header, gw.count_rows() + 1):
-            row_values = values[row-1]
-            url = gw.get_cell(row_values, 'url')
-            status = gw.get_cell(row_values, 'status')
+            url = gw.get_cell(row, 'url')
```
msramalho, line 110: only 1 call is made to get_values, so we can do as much get_row/get_cell as needed

loganwilliams: This is almost perfect, except that it was intentional that the last check of the status cell, just before starting the archive, does a new API call to the sheet. This is because there can be multiple archivers operating on a sheet simultaneously (in the case of capturing live streaming content especially).

To recap: if the "cached" status is non-blank, skip it. If the "cached" status is blank, check it again with a live API call, and then archive if it is still blank.

Make sense?

archivers/telegram_archiver.py
loganwilliams: :100:

auto_archive.py
loganwilliams: This is almost perfect, except that it was intentional that the last check of the status cell, just before starting the archive, does a new API call to the sheet. This is because there can be multiple archivers operating on a sheet simultaneously (in the case of capturing live streaming content especially).

To recap: if the "cached" status is non-blank, skip it. If the "cached" status is blank, check it again with a live API call, and then archive if it is still blank.

Make sense?

loganwilliams (changes requested):

msramalho:
Thanks for the feedback, last commit addresses the issue raised.

PR #21, opened 2022-03-18
Resolve telethon issues

loganwilliams:
- [x] Selenium driver not working for screenshots (however this fix uses a `FirefoxProfile` which has apparently been deprecated, so it would be nice to find a non-deprecated way to do this)
- [x] Message comparison weirdness, int vs str (fixed by adding another len, but not sure if this is the intended logic
- [x] Garbled Cyrillic text in HTML output -- possibly a character encoding issue when something is upload to S3
- [x] `post` can be None? This throws the error quoted below:

```
2022-03-18 11:06:22.340 | ERROR    | __main__:process_sheet:134 - Got unexpected error in row 2879 with archiver TelethonArchiver for url [...]: 'NoneType' object has no attribute 'grouped_id'
Traceback (most recent call last):
  File "/Volumes/Oxygen/auto-archiver/auto_archive.py", line 131, in process_sheet
    result = archiver.download(url, check_if_exists=True)
  File "/Volumes/Oxygen/auto-archiver/archivers/telethon_archiver.py", line 65, in download
    media_posts = self._get_media_posts_in_group(chat, post)
  File "/Volumes/Oxygen/auto-archiver/archivers/telethon_archiver.py", line 41, in _get_media_posts_in_group
    if post.grouped_id == original_post.grouped_id and post.media is not None:
AttributeError: 'NoneType' object has no attribute 'grouped_id'
```

### DIFF
--- a/archivers/telethon_archiver.py
+++ b/archivers/telethon_archiver.py
@@ -37,7 +37,7 @@ class TelethonArchiver(Archiver):
         posts = self.client.get_messages(chat, ids=search_ids)
         media = []
         for post in posts:
-            if post.grouped_id == original_post.grouped_id and post.media is not None:
+            if post is not None and post.grouped_id == original_post.grouped_id and post.media is not None:
                 media.append(post)
         return media
 
--- a/archivers/telethon_archiver.py
+++ b/archivers/telethon_archiver.py
@@ -2,7 +2,6 @@ import os
 import re
 import html
 from dataclasses import dataclass
-from urllib.parse import urlparse
 from loguru import logger
 
 from storages import Storage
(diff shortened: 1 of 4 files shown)


msramalho (approved):

msramalho:
approved so far, can look at other issues afterwards

archivers/telethon_archiver.py
```diff
@@ -76,7 +76,7 @@ def download(self, url, check_if_exists=False):
                 uploaded_media = []
                 message = post.message
                 for mp in media_posts:
-                    if len(mp.message) > message: message = mp.message
+                    if len(mp.message) > len(message): message = mp.message
```
msramalho, line 79: this is the correct version

msramalho:
Fixed all issues reported.

PR #28, opened 2022-05-09
youtubedl_archiver tweak for twitter.com to see if a linked video is there

djhmateer:
…video is in there eg vk.com

For example in this Tweet

https://twitter.com/RALee85/status/1496790012242837506

There is a linked video to https://vk.com/milinfolive?w=wall-123538639_2215573

The assumption is the user meant the tweet and corresponding 4 images to be archived and not the linked video.

A Twitter embedded video https://twitter.com/FreeBurmaRangrs/status/1500900146216943621 will work as comes from twitter domain.


msramalho:
This makes sense, especially with the current order where the TwitterArchiver comes afterwards.

msramalho (approved):

PR #29, opened 2022-05-09
Load Facebook cookie from .env file

djhmateer:
To make deployments easier I've put the FB cookie into the .env file.


msramalho (approved):
Useful cleanup!

PR #30, opened 2022-05-09
Add Logging output to logs folder

djhmateer:
To help in production I've added in these levels of logging, and the 'catch all' decorator on main.

I've tried to adhere to standard logging levels:

https://github.com/djhmateer/djhmateer.github.io/blob/master/_posts/2022-04-25-loguru.md

And put in the top level catch all which is useful when Google API sometimes doesn't work querying the Google Sheet

### DIFF
--- a/auto_archive.py
+++ b/auto_archive.py
@@ -20,6 +20,7 @@ logger.add("logs/3success.log", level="SUCCESS")
 logger.add("logs/4warning.log", level="WARNING")
 logger.add("logs/5error.log", level="ERROR")
 
+load_dotenv()
 
 def update_sheet(gw, row, result: archivers.ArchiveResult):
     cell_updates = []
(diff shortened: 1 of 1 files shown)


auto_archive.py
```diff
@@ -86,7 +92,7 @@ def process_sheet(sheet, header=1, columns=GWorksheet.COLUMN_NAMES):
 
     # loop through worksheets to check
     for ii, wks in enumerate(sh.worksheets()):
-        logger.info(f'Opening worksheet {ii}: "{wks.title}" header={header}')
+        logger.info(f'Opening worksheet ii={ii}: {wks.title} header={header}')
```
msramalho, line 95:
```suggestion
        logger.info(f'Opening worksheet {ii=}: {wks.title=} {header=}')
```
this will get you the same result, see [py3.8 release notes](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging)

PR #35, opened 2022-05-11
Folders and Google Drive

djhmateer:
Implementations of:

- Save into a folder on cloud storage from a column in the spreadsheet called File Number. Works with S3 (Digital Ocean Spaces) and GD (Google Drive). In fact it is mandatory for GD

- Save to Google Drive storage instead of DO.

This was coded to be as simple as I could make it. There is a lot of room for refactoring!

A big change was the Google Drive works on ID's, not on relative file paths and names. Also GD doesn't allow you to serve a static html page, which caused the thumbnails page to break. This was not a problem for us.

As far as my testing has gone, I believe this to all work as expected, and it shouldn't break any of your sheets. For example the new `File Number` column is not required.


djhmateer:
And apologies for such a beast of a PR.

msramalho:
Hi @djhmateer I've refactored the whole code a bit since the `filenumber` config provoked a lot of changes in a not-easy to maintain way. I will merge it to `dev` branch now, then will work my other PR into it, do further cleanups and merge to master.
When I merge to master I will explain how the `filenumber` -> `subfolder` changes works, but the idea is that any storage can use it if available, and if it is missing then GDarchiver will default to a folder `default`. 

Thank you very much for figuring out how to do this.

djhmateer:
Hi @msramalho many thanks for merging in to dev and for refactoring. I look forward to seeing other changes and merge up into master. Let me know if I can help.

PR #55, opened 2022-07-18
Put in hash_algorithm in config to switch between SHA256 and SHA3_512

djhmateer:
Would appreciated refactoring's as my knowledge of Python OO is limited. This is very useful for us and having in a config allows much easier testing and deployment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ config-*.yaml
 logs/*
 local_archive/
 vk_config*.json
+
+secrets/*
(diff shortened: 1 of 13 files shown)


msramalho:
Hi @djhmateer I've updated the names so they are standard instead of reflecting python's nomenclature `SHA-256 or SHA3-512` and made this into a global Archiver property like TMP_FOLDER is for Storage, as this way it requires less visual complexity, there could be better OO approaches here, but those would require some refactor that I don't think is worth the time. Let me know if you see any issues, otherwise I'll merge to dev soon.

PR #56, opened 2022-07-18
Oauth integration with Google Drive so can use a real user and not a service account

djhmateer:
Have been using this code for a month and it seems good. Token refreshes are handled.

Also updated a 'trashed = false' query for google drive so nothing in the recycle bin of google drive is picked up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,6 @@ config-*.yaml
 logs/*
 local_archive/
 vk_config*.json
-
+gd-token.json
+credentials.json
 secrets/*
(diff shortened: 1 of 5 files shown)


msramalho:
Can you share a link where the 15GBs limit is documented? is this per a fixed time window or in total?

msramalho:
NOTE: renamed `oauth_token_file_path_and_name` to `oauth_token_filename`

msramalho:
ready in dev, I've tested it and works as expected, added clarification about the 15GB limit, that is because of a personal account, the token allows the storage to count towards the owner of the folder vs the developer account

PR #57, opened 2022-07-19
Empty entry number if using Google Storage catch

djhmateer:
A small change so that if using Google Drive and there is an empty entry number, which correlates to the new folder be created, it is ignored until the next run.

This is a 'normal' workflow under some conditions.

### DIFF
--- a/auto_archive.py
+++ b/auto_archive.py
@@ -93,6 +93,14 @@ def process_sheet(c: Config):
                 is_retry = Archiver.should_retry_from_status(status)
                 if not is_retry: continue
 
+            # if using folder based storage make sure an folder/Entry Number is specified eg MW0001
+            if c.storage == 'gd':
+                entry_number = gw.get_cell(row, 'folder')
+                # Some Slack integration writes a URL before an Entry Number is written so just leave for next time
+                if entry_number == "": 
+                    logger.warning(f'Missing entry number - waiting for next run')
+                    continue
+
             # All checks done - archival process starts here
             try:
                 gw.set_cell(row, 'status', 'Archive in progress')


msramalho:
Hi Dave, I was looking into this logic and I don't think it makes sense to include it in the base-code since it is very specific to your project. We have a default folder value (controllable in: https://github.com/bellingcat/auto-archiver/blob/3095ce305459b40bd60d73ba6ee84d834449fd21/auto_archive.py#L86) but making an exception that every google drive archival that does not have a folder will be skipped would break other users' expectations.
One suggestion is to handle this in your google sheets, for example:
1. you get a URL in a column unconfirmedURL
2. when you get your entry number from slack the actual URL column will detect it and copy the value from unconfirmedUrl, you could do that with sth like `=IF(LEN(entryNumberCell)==0, "", unconfirmedUrlCell)`
3. this way the archiver which should only look for the URL col would only see it (and therefore archive) when the entryNumber is present


Let me know if this made sense to your use case, because I only deduced it from the comments.

djhmateer:
Agreed this is a very specific edge case.

Thanks for thinking this through. I'd prefer not to change the sheets now (as there are quite a few).

Would it be acceptable to create a new config under `google_drive` called `ignore_if_folder_not_specified_in_spreadsheet`

If this is too 'edgey' then I'm totally fine with  just keeping it in mine.

Thanks again for all your work.

msramalho:
Hi Dave,
Indeed I think that will be to edge-casey as the normal flow creates/uses the default folder names that come from the `spreadsheet name/worksheet name`, thanks for understanding.

PR #93, opened 2023-09-12
Add configs for path to write and read wacz archives

liliakai:
To support running browstrix in docker while also running the auto-archiver(api) in docker, we add separate configs for where browsertrix writes the archives and where the wacz enricher reads the archives, because one is relative to the host and one is relative to the container, even though they may refer to the same volume.


liliakai:
Perhpas these configs should be environment variables instead, so one could set them directly from a docker compose file, reducing the amount of configuration needed. What do you think @msramalho ?

msramalho (approved):

PR #106, opened 2023-11-07
Fix Selenium headless invokation

### LINKED ISSUE #103: Selenium headless not working - workaround to force 4.12.0
Have found that when I did a `pipenv update` then ran the code on my local machine, the firefox browser would display briefly. However on the server (with no terminal attached) it couldn't, so errored out, not producing a screenshot

In my Pipfile I've added this for the meantime which works as before
```
#selenium = "*"
# bug where it renders to screen on 4.15.0, and fails on the server
selenium = "==4.12.0"
```

GalenReich:
Closes #103 

Selenium has removed support for the headless option, this PR changes the archiver to use the new `--headless` argument.

They talk about the change [on their blog](https://www.selenium.dev/blog/2023/headless-is-going-away/), and made the change in [this commit](https://github.com/seleniumhq/selenium/commit/b5cfcc4f9c51dd9df9106b62971c417a0fc9a5e4).

This should be tested before merging.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,96 +18,96 @@
     "default": {
         "aiohttp": {
             "hashes": [
-                "sha256:00ad4b6f185ec67f3e6562e8a1d2b69660be43070bd0ef6fcec5211154c7df67",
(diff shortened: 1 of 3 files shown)


msramalho (approved):
Thanks for the fix and detailed reference materials!

Issue #98, opened 2023-09-18
no _mimetype set for final media html files


liliakai:
A small gripe, but nice to have for consistency.

Using config `formatter: html_formatter`, the final media on an archive is an html page generated by the orchestrator. In practice I find that other media objects have a correctly set `_mimetype` property, but not the one from html_formatter. Seems like [core/media.py](https://github.com/bellingcat/auto-archiver/blob/e8935b9a805a9487271ba76b84730edb940046a4/src/auto_archiver/core/media.py#L70) should set the right mimetype based on filename. I think it just never gets called for the final media.

msramalho:
these lines https://github.com/bellingcat/auto-archiver/blob/9eb39943c7c99a6316ff8d1c9ae3371377736c5e/src/auto_archiver/storages/s3.py#L54-L59 would do that for S3, local/google-drive storage is less relevant, unless you mean mimetype somewhere else?

And it gets set on S3 stored metadata objects.

liliakai:
I mean in the metadata stored in the db rather than s3

PR #111, opened 2023-12-08
Set _mimetype for final media of html formatter

JettChenT:
#98


msramalho (approved):

msramalho:
Thanks for the contribution!

PR #112, opened 2023-12-12
reduce s3 duplicating while keeping random urls via hash

msramalho:


msramalho:
adds new s3 configuration to allow configuring the described behaviour

PR #123, opened 2024-01-27
fix: Correct the path of service account in google drive settings

andyfcx:
The functionality using GDrive with service account is not working, due to loading config from the yaml is not right.

### DIFF
--- a/src/auto_archiver/storages/gd.py
+++ b/src/auto_archiver/storages/gd.py
@@ -87,15 +87,6 @@ class GDriveStorage(Storage):
         file_id = self._get_id_from_parent_and_name(folder_id, filename)
         return f"https://drive.google.com/file/d/{file_id}/view?usp=sharing"
 
-    # def upload(self, media: Media, **kwargs) -> bool:
-    #     # override parent so that we can use shutil.copy2 and keep metadata
-    #     dest = os.path.join(self.save_to, media.key)
-    #     os.makedirs(os.path.dirname(dest), exist_ok=True)
-    #     logger.debug(f'[{self.__class__.name}] storing file {media.filename} with key {media.key} to {dest}')
-    #     res = shutil.copy2(media.filename, dest)
-    #     logger.info(res)
-    #     return True
-
     def upload(self, media: Media, **kwargs) -> bool:
         logger.debug(f'[{self.__class__.name}] storing file {media.filename} with key {media.key}')
         """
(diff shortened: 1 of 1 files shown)


djhmateer:
Hi @andyfcx  On a tangent.. If you're using the service account strategy for Google Drive (which I did for a while), watch out for the 15GB limit. I made some notes here: https://davemateer.com/2022/04/28/google-drive-with-python

I switched over to OAuth.

Thanks for submitting a PR!

src/auto_archiver/storages/gd.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, config: dict) -> None:
             else:
                 logger.debug('GD OAuth Token valid')
         else:
-            gd_service_account = config.service_account
+            gd_service_account = config['gdrive_storage']['service_account']
```
msramalho, line 55:
```suggestion
            gd_service_account = self.service_account
```
This is actually the standard way of accessing the `configs` via `self`.

src/auto_archiver/storages/gd.py
```diff
@@ -87,15 +87,6 @@ def get_cdn_url(self, media: Media) -> str:
         file_id = self._get_id_from_parent_and_name(folder_id, filename)
         return f"https://drive.google.com/file/d/{file_id}/view?usp=sharing"
 
-    def upload(self, media: Media, **kwargs) -> bool:
```
msramalho, line 90: great catch, this is the local_storage code.

src/auto_archiver/storages/gd.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, config: dict) -> None:
             else:
                 logger.debug('GD OAuth Token valid')
         else:
-            gd_service_account = config.service_account
+            gd_service_account = self.service_account
```
msramalho, line 55: I've fixed the referencing here

msramalho (changes requested):
only pending thing is explaining line 153, otherwise ready to go.

src/auto_archiver/storages/gd.py
```diff
@@ -159,6 +150,8 @@ def _get_id_from_parent_and_name(self, parent_id: str, name: str, retries: int =
             ).execute()
             items = results.get('files', [])
 
+            items = self.service.ListFile(query_string).GetList()
```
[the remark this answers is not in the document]
andyfcx: Ah, sorry, this line is unnecessary.

msramalho:
code is released in 0.8.1: https://pypi.org/project/auto-archiver/

knigady:
gd.py cannot find the folder on Google Drive with the ID specified in the orchestration.yaml file, and as a result, it sets the value to None and uploads the file to the root of the drive. What could this be related to?

PR #129, opened 2024-02-14
closes several open issues: new enrichers and bug fixes

msramalho:


msramalho:
there's some duplication of code between telethon_archiver and instagram_bot_archiver but the code isolation had a low cost/benefit for the moment

PR #138, opened 2024-04-03
Add yt-dlp based archiving for TwitterArchiver

JettChenT:
This PR adds a yt-dlp based twitter archiving function to `TwitterArchiver` as a fallback to the existing two archiving strategies. It uses the `_extract_status` function of yt-dlp's `TwitterIE` extractor to extract tweet metadata, and processes it in a similar way to the existing archiving implementation.

Upon local testing, the existing snscrape(which seems to be unmaintained) and twitter-hack solution does not work reliably for tweets, but the yt-dlp based solution does. Happy to know if it can be replicated! 

Also happy to add a configuration option to `TwitterArchiver` for specifying the preference of tweet archiving methods(`snscrape`/`twitter-hack`/`yt-dlp`)


src/auto_archiver/archivers/twitter_archiver.py
```diff
@@ -140,13 +176,13 @@ def choose_variant(self, variants):
...
+                width_height = re.search(r"\/(\d+)x(\d+)\/", var["url"])
                 if width_height:
                     w, h = int(width_height[1]), int(width_height[2])
                     if w > width or h > height:
                         width, height = w, h
-                        variant = var.get("src", variant)
+                        variant = var
             else:
-                variant = var.get("src") if not variant else variant
+                variant = var if not variant else variant
```
msramalho, line 187: these key changes in variant_choosing may break the other scrapers, but as they are well... broken ... it's just fine for now.

msramalho (approved):
This is a really good and impactful contribution, thank you for taking the time to properly implementing this new alternative. As we're building on 3 and ytdlp is the most likely to work now, I would accept refactoring the priorities of the code but that's an extra and makes little impact in the running time. will merge

PR #141, opened 2024-04-20
twitter_archiver timeout on sanitize url and added try except 

djhmateer:
A few tweaks.

### DIFF
--- a/src/auto_archiver/archivers/twitter_archiver.py
+++ b/src/auto_archiver/archivers/twitter_archiver.py
@@ -31,7 +31,8 @@ class TwitterArchiver(Archiver):
         # expand URL if t.co and clean tracker GET params
         if 'https://t.co/' in url:
             try:
-                r = requests.get(url)
+                # timeout for urls like https://t.co/KxNwn5yJz6 which caused this request to never come back
+                r = requests.get(url, timeout=30)
                 logger.debug(f'Expanded url {url} to {r.url}')
                 url = r.url
             except:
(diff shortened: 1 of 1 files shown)


src/auto_archiver/archivers/twitter_archiver.py
```diff
@@ -31,7 +31,8 @@ def sanitize_url(self, url: str) -> str:
         # expand URL if t.co and clean tracker GET params
         if 'https://t.co/' in url:
             try:
-                r = requests.get(url)
+                # timeout for urls like https://t.co/KxNwn5yJz6 which caused this request to never come back
+                r = requests.get(url, timeout=30)
```
msramalho, line 35: this makes sense

src/auto_archiver/archivers/twitter_archiver.py
```diff
@@ -52,11 +53,21 @@ def download(self, item: Metadata) -> Metadata:
         result = Metadata()
 
         scr = TwitterTweetScraper(tweet_id)
+        do_alternative = False
```
msramalho, line 56: What is the goal of this hardcoded value? It will never trigger the condition below.

msramalho:
Thanks for bringing attention to this code.
I think I'll just refactor the entire file logic as it's now doing the attempts in a nested way and not in the best (as in most likely to work) order. I'll include some of the changes you mentioned.

msramalho:
Done, please see https://github.com/bellingcat/auto-archiver/pull/142

closing this PR

djhmateer:
Thanks Miguel

PR #153, opened 2024-11-25
Optimize the Docker file - fixes #1

### LINKED ISSUE #1: Project should be easier to set up and run
Currently, running this project in an automated way requires creating a Digital Ocean Spaces bucket and manually managing cron jobs on a Linux server. Ideally, this would be simpler to deploy so that a new archiving spreadsheet could be set up in a user friendly way, even for non-programmers.

One promising possibility for moving in this direction is as a Google Sheets Add On, but other ideas can also be explored and evaluated.

darknetehf:
Hello,

This is an attempt to optimize the Docker file.
I noticed that the base image webrecorder/browsertrix-crawler weighs 2.44 Gb, which is already a lot, whereas the resulting auto-archiver image nearly doubles the size (4.29GB on my end). There is clearly some bloat.

I think we could improve this with multi-stage builds and remove some unwanted layers.

The proposed version should be functionally equivalent. The image size is 3.33GB on my end.
There is one stage that builds the virtual environment and copies it to the next layer. So, we can ditch pipenv and run Python directly. See the adapted entry point.

Another benefit of the multi-stage build is that if changes are made to the Python code, there is no need to rebuild the upper layers unless the requirements list has changed. Thanks to caching, generating a new Docker image will be quicker.

However, I need to stress that it’s not been tested thoroughly at all, and should be treated as a PoC to be validated before going to production.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,30 +1,44 @@
-FROM webrecorder/browsertrix-crawler:1.0.4
+FROM webrecorder/browsertrix-crawler:1.0.4 AS base
 
-ENV RUNNING_IN_DOCKER=1
-
-WORKDIR /app
-
-RUN pip install --upgrade pip && \
-	pip install pipenv && \
-	add-apt-repository ppa:mozillateam/ppa && \
+RUN add-apt-repository ppa:mozillateam/ppa && \
 	apt-get update && \
-	apt-get install -y gcc ffmpeg fonts-noto exiftool && \
+	apt-get install -y --no-install-recommends gcc ffmpeg fonts-noto exiftool && \
 	apt-get install -y --no-install-recommends firefox-esr && \
 	ln -s /usr/bin/firefox-esr /usr/bin/firefox && \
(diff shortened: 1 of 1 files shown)


msramalho:
thanks for the submission! these are relevant changes.
as we're moving to poetry and that will already have a significant impact on the docker image size, let's keep this on hold until then and reassess.

pjrobertson:
Many of these changes have been merged into main as part of #164 , and the current docker file is now 3.61GB (for ARM). That's ~270MB larger than this package here, but the latest version of auto-archiver also includes several other packages (ruamel-yaml, opentimestamps etc.) and it's possible the browsertrix package has also changed in size.

Thanks darknetehf for this. I think we are good to go with the current dockerfile, but if there are any further optimisations that can be done, then please do submit another PR from the latest BASE using poetry!

P.S. I don't know much about the advantages/disadvantages of `PYTHONDONTWRITEBYTECODE` , but https://aleksac.me/blog/dont-use-pythondontwritebytecode-in-your-dockerfiles/ seems to suggest it's better to avoid doing that.

PR #157, opened 2024-12-03
Make HashEnricher config optional

### LINKED ISSUE #156: No hashing algorithm defined for use in HtmlFormatter when HashEnricher not included
Given that `HtmlFormatter` checks `ArchivingContext` for the appropriate hashing algorithm, [that fails](https://github.com/user-attachments/files/17996014/no_hashing_algorithm.log) when `HashEnricher` is not included in `orchestration.yaml`.

https://github.com/bellingcat/auto-archiver/blob/e49550163f1fb5a704a1b69c4aa6ff9ca3cdef26/src/auto_archiver/formatters/html_formatter.py#L57-L59

In such a scenario, as `HashEnricher` is not initiated earlier, `ArchivingContext` has no record of a selected algorithm (default or otherwise).

maxtheaxe:
Fixes #156. The same concept could be applied to `Step` itself, by updating `__init__()` with something like the following.
```
if config:
    for k, v in config.get(self.name, {}).items():
        self.__setattr__(k, v)
else: # allow initialization with default values
    for k, v in self.configs():
        self.__setattr__(k, v)
```
I wasn't sure of your preferences, so I figured I'd only add it where necessary for the time being, though I'm happy to refactor it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ auto_archiver.egg-info*
 logs*
 *.csv
 archived/
+.idea/
--- a/src/auto_archiver/enrichers/hash_enricher.py
+++ b/src/auto_archiver/enrichers/hash_enricher.py
@@ -11,9 +11,20 @@ class HashEnricher(Enricher):
     """
     name = "hash_enricher"
 
-    def __init__(self, config: dict) -> None:
+    def __init__(self, config: dict = None) -> None:
+        final_config = {}
+        if not config:
+            # currently doesn't use user config in scenarios in which hash_enricher
+            # config is defined in yaml, but hash_enricher itself is not (as config
+            # parsing happens elsewhere)
+            he_configs = self.configs()
+            # handle potential future config options
+            for property_name in he_configs:
+                final_config[property_name] = he_configs[property_name]["default"]
+        else:
+            final_config = config
         # without this STEP.__init__ is not called
-        super().__init__(config)
+        super().__init__({self.name: final_config})
         algo_choices = self.configs()["algorithm"]["choices"]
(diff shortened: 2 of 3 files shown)


msramalho:
thanks for the submission! The root issue here is indeed having the html_formatter depend on the hash_enricher being used -> not a guarantee. My preference will be to actually have a sensible default value here

https://github.com/bellingcat/auto-archiver/blob/feat/bluesky-archiver/src/auto_archiver/formatters/html_formatter.py#L57

rather than (maintaing the) coupling the hash_enricher with other steps' logic. 

we may also need to rethink the need for the archivingContext altogether here, the "original sin" was blindly coupling both steps in the past, so we should revert that coupling instead.

pjrobertson:
Thanks for the contributions, @maxtheaxe ! We are working on a rewrite of the auto-archiver structure, which no longer uses `ArchivingContext`. Once the new version is released, I'm happy to say that the HashEnricher used in the `html_formatter` will work with the default values of the `hash_enricher`, even if it's not enabled as an enricher. This should also fix #156.

For reference, the relevant changes/mechanism that enables this is that modules (e.g. the html_formatter) can now have 'dependency' modules which are also loaded on startup, with either the config info or with the default values.
The html_formatter is set to be dependent on the hash_enricher [here](https://github.com/bellingcat/auto-archiver/blob/91ca325fd510c0c1989b6b31050571331a074c3b/src/auto_archiver/modules/html_formatter/__manifest__.py#L6). Which means the dependency is first loaded when the html_formatter is being loaded [here](https://github.com/bellingcat/auto-archiver/blob/91ca325fd510c0c1989b6b31050571331a074c3b/src/auto_archiver/core/module.py#L196)

I'll close this PR for now - keep your eyes on the project for an update coming hopefully very soon!  👀

PR #158, opened 2024-12-20
Allow setting cookies for yt-dl

pjrobertson:
Settings can be adjusted in orchestration.yml, with one of two options:

```
    # use one of the following two methods to authenticate in youtube - either provide a cookies file or use the cookies of the given browser
    # for more information, see https://github.com/yt-dlp/yt-dlp/wiki/FAQ#how-do-i-pass-cookies-to-yt-dlp
    # cookie_file: "secrets/youtube_cookies.txt"
    # cookies_from_browser: firefox
```

Todo: create a PR template, to ensure the following are done:
1. Tests passing
2. Documentation updated [need documentation guideline - do we put documentation in orchestration.yml or in command line args config description?
3. Reviewer (at least 1 other person)

### DIFF
--- a/example.orchestration.yaml
+++ b/example.orchestration.yaml
@@ -94,6 +94,14 @@ configurations:
     password: "vk pass"
     session_file: "secrets/vk_config.v2.json"
 
+  youtubedl_archiver:
+    subtitles: true
+    # use one of the following two methods to authenticate in youtube - either provide a cookies file or use the cookies of the given browser
+    # for more information, see https://github.com/yt-dlp/yt-dlp/wiki/FAQ#how-do-i-pass-cookies-to-yt-dlp
+    # cookie_file: "secrets/youtube_cookies.txt"
+    # cookies_from_browser: firefox
+    # proxy: socks5://proxy-user:password@proxy-ip:port
+
   screenshot_enricher:
     width: 1280
     height: 2300
--- a/src/auto_archiver/archivers/youtubedl_archiver.py
+++ b/src/auto_archiver/archivers/youtubedl_archiver.py
@@ -30,6 +30,8 @@ class YoutubeDLArchiver(Archiver):
             "end_means_success": {"default": True, "help": "if True, any archived content will mean a 'success', if False this archiver will not return a 'success' stage; this is useful for cases when the yt-dlp will archive a video but ignore other types of content like images or text only pages that the subsequent archivers can retrieve."},
             'allow_playlist': {"default": False, "help": "If True will also download playlists, set to False if the expectation is to download a single video."},
             "max_downloads": {"default": "inf", "help": "Use to limit the number of videos to download when a channel or long page is being extracted. 'inf' means no limit."},
+            "cookies_from_browser": {"default": None, "help": "optional browser for ytdl to extract cookies from, can be one of: brave, chrome, chromium, edge, firefox, opera, safari, vivaldi, whale"},
+            "cookie_file": {"default": None, "help": "optional cookie file to use for Youtube, see instructions here on how to export from your browser: https://github.com/yt-dlp/yt-dlp/wiki/FAQ#how-do-i-pass-cookies-to-yt-dlp"},
         }
 
     def download(self, item: Metadata) -> Metadata:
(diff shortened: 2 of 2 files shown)


src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -30,6 +30,8 @@ def configs() -> dict:
             "end_means_success": {"default": True, "help": "if True, any archived content will mean a 'success', if False this archiver will not return a 'success' stage; this is useful for cases when the yt-dlp will archive a video but ignore other types of content like images or text only pages that the subsequent archivers can retrieve."},
             'allow_playlist': {"default": False, "help": "If True will also download playlists, set to False if the expectation is to download a single video."},
             "max_downloads": {"default": "inf", "help": "Use to limit the number of videos to download when a channel or long page is being extracted. 'inf' means no limit."},
+            "cookies_from_browser": {"default": None, "help": "optional browser for ytdl to extract cookies from, can be one of: brave, chrome, chromium, edge, firefox, opera, safari, vivaldi, whale"},
```
msramalho, line 33: there is a `"choices": ['...']` key for the config values but as it stands I don't think it allows for `None` to be selected, so this is just all right but a small improvement to consider in the more general config.py logic.

example.orchestration.yaml
```diff
@@ -94,6 +94,14 @@ configurations:
...
     session_file: "secrets/vk_config.v2.json"
 
+  youtubedl_archiver:
+    subtitles: true
+    # use one of the following two methods to authenticate in youtube - either provide a cookies file or use the cookies of the given browser
+    # for more information, see https://github.com/yt-dlp/yt-dlp/wiki/FAQ#how-do-i-pass-cookies-to-yt-dlp
+    # cookie_file: "secrets/youtube_cookies.txt"
+    # cookies_from_browser: firefox
+    # proxy: socks5://proxy-user:password@proxy-ip:port
+
```
msramalho, line 104: It's fine to leave these here for now, one interesting improvement we can work on is to auto generate the example file based on the Configurations of a given Step.

msramalho (approved):
Looks good to merge! 

I commented a lot to show some of my thinking around the existing code itself.
In fact, it actually made me reflect on how we are currently hardcoding a 1-1 mapping of yt-dlp command line arguments when in fact we could have a more general way of allowing those to come directly as a single Configuration, happy to discuss that idea further. some inspiration: https://github.com/yt-dlp/yt-dlp/issues/5859#issuecomment-1363938900

src/auto_archiver/archivers/youtubedl_archiver.py
pjrobertson: True. yt-dlp also allows tuples for  (`browsername:profilename`) but I thought that might be a bit too complicated to include here (for the average user). Happy to extend this to allow either `browser` or `browser:profilename` if you think it is worth it.

Ref: https://github.com/yt-dlp/yt-dlp/wiki/FAQ#how-do-i-pass-cookies-to-yt-dlp

pjrobertson:
> when in fact we could have a more general way of allowing those to come directly as a single Configuration, happy to discuss that idea further. some inspiration: [yt-dlp/yt-dlp#5859 (comment)](https://github.com/yt-dlp/yt-dlp/issues/5859#issuecomment-1363938900)

I like this idea. Could be a configuration like `yt-dlp` and then you can just apply a text string like `cookies-from-browser chrome --other-argument value`.

I still think we should make the simplest/basic arguments available in the config file to the end user, so that they don't have to try and work out what the command line options are. I think cookies is one that is 'higher level'. For example, I was  that the 'cookies' settings should actually be abstracted out from the `youtubedl_archiver` settings and made global. Then the cookies can be used for all archivers (e.g. we already have a facebook cookie setting, it would be better to provide a single cookies file/setting that can be used for this and other archivers).

example.orchestration.yaml
pjrobertson: Yes, Erin also mentioned in the bluesky PR about the expectations for editing example.orchestration.yaml. We should come up with guidelines. I can discuss with E further and we can create some initial guidelines.

msramalho: please do!

msramalho:
definitely worth opening an issue/thread to rethink how cookies are fed into multiple archivers.

PR #160, opened 2024-12-23
adds an unauthenticated Bluesky archiver

msramalho:
- avoids ATProto
- leaves a few TODOs in reusable download_from_url method

### DIFF
--- a/example.orchestration.yaml
+++ b/example.orchestration.yaml
@@ -2,6 +2,7 @@ steps:
   # only 1 feeder allowed
   feeder: gsheet_feeder # defaults to cli_feeder
   archivers: # order matters, uncomment to activate
+    - bluesky_archiver
     # - vk_archiver
     # - telethon_archiver
     # - telegram_archiver
--- a/src/auto_archiver/archivers/__init__.py
+++ b/src/auto_archiver/archivers/__init__.py
@@ -9,3 +9,4 @@ from .telegram_archiver import TelegramArchiver
 from .vk_archiver import VkArchiver
 from .youtubedl_archiver import YoutubeDLArchiver
 from .instagram_api_archiver import InstagramAPIArchiver
+from .bluesky_archiver import BlueskyArchiver
--- a/src/auto_archiver/archivers/archiver.py
+++ b/src/auto_archiver/archivers/archiver.py
@@ -48,6 +48,8 @@ class Archiver(Step):
         """
         downloads a URL to provided filename, or inferred from URL, returns local filename
         """
+        # TODO: should we refactor to use requests.get(url, stream=True) and write to file in chunks? compare approaches
+        # TODO: should we guess the extension?
         if not to_filename:
             to_filename = url.split('/')[-1].split('?')[0]
             if len(to_filename) > 64:
--- a/src/auto_archiver/archivers/bluesky_archiver.py
+++ b/src/auto_archiver/archivers/bluesky_archiver.py
@@ -1,0 +1,117 @@
+import os
+import re, requests, mimetypes
+from loguru import logger
+
+
+from . import Archiver
+from ..core import Metadata, Media, ArchivingContext
+
+
+class BlueskyArchiver(Archiver):
+    """
+    Uses an unauthenticated Bluesky API to archive posts including metadata, images and videos. Relies on `public.api.bsky.app/xrpc` and `bsky.social/xrpc`. Avoids ATProto to avoid auth.
+
+    Some inspiration from https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/extractor/bluesky.py
+    """
+    name = "bluesky_archiver"
+    BSKY_POST = re.compile(r"/profile/([^/]+)/post/([a-zA-Z0-9]+)")
+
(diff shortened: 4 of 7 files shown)


src/auto_archiver/archivers/__init__.py
```diff
@@ -8,4 +8,5 @@
 from .telegram_archiver import TelegramArchiver
 from .vk_archiver import VkArchiver
 from .youtubedl_archiver import YoutubeDLArchiver
-from .instagram_api_archiver import InstagramAPIArchiver
\ No newline at end of file
+from .instagram_api_archiver import InstagramAPIArchiver
+from .bluesky_archiver import BlueskyArchiver
```
pjrobertson, line 12: Just a general comment here (not relevant to this PR) but I don't like how we import *all* archivers in the __init__.py file. Ideally, we should only load dependencies (archivers, feeders etc.) that are enabled in the orchestration.

Reason: I came across a bug whereby an archiver was broken (snscrape was broken), and even though I hadn't enabled it, it was still being loaded and throwing an error. Imagine if an archiver breaks for whatever reason (or more likely, an archiver's dependencies break), it would be much easier to just disable the archiver than telling the user to find an `__init__.py` file and 'change the python code'.

This is essentially how django works. You create a list of `INSTALLED_APPS` and then the archiver tries to install them one by one. We already have this list of 'installed apps' (it's called steps in our orchestration). So it's just a matter of on startup, trying to load each of them. Then when they can't be loaded throwing an error.

pjrobertson:
LGTM. Tested and works.

One issue I found was the media wasn't getting downloaded for this url: https://bsky.app/profile/colborne.bsky.social/post/3lec2bqjc5s2y

For some reason, bsky uses 2 different places to store images. Under `images` and also under `media/images` (I think the latter is if there are multiple images)

```
(Pdb) post.get("record", {}).get("embed", {})
{'$type': 'app.bsky.embed.recordWithMedia', 'media': {'$type': 'app.bsky.embed.images', 'images': [{'alt': '', 'aspectRatio': {'height': 1119, 'width': 922}, 'image': {'$type': 'blob', 'ref': {'$link': 'bafkreiflrkfihcvwlhka5tb2opw2qog6gfvywsdzdlibveys2acozh75tq'}, 'mimeType': 'image/jpeg', 'size': 479309}}, {'alt': '', 'aspectRatio': {'height': 1333, 'width': 2000}, 'image': {'$type': 'blob', 'ref': {'$link': 'bafkreibsprmwchf7r6xcstqkdvvuj3ijw7efciw7l3y4crxr4cmynseo7u'}, 'mimeType': 'image/jpeg', 'size': 800522}}]}, 'record': {'$type': 'app.bsky.embed.record', 'record': {'cid': 'bafyreiavgtdninmcqeoiwrw7pgjubgu7rjfdd7jfb7dwvbvj7aavppwkxm', 'uri': 'at://did:plc:pr46lkz3rkp2yvxnvwe47ib2/app.bsky.feed.post/3ldpxkzv53k2q'}}}
```

Looking at the [bksy source](https://github.com/bluesky-social/atproto/blob/main/packages/bsky/src/lexicon/types/app/bsky/feed/defs.ts#L18), it seems that embed can be one of:

```
  embed?:
    | AppBskyEmbedImages.View
    | AppBskyEmbedVideo.View
    | AppBskyEmbedExternal.View
    | AppBskyEmbedRecord.View
    | AppBskyEmbedRecordWithMedia.View
    | { $type: string; [k: string]: unknown }
```

I've modified the code to look at both these paths for images. I also took the liberty of adding in some unit tests to test the 3 different types of media (single image, multiple image, video).

The unit tests might want to be abstracted out to another PR. But it's a framework on how unit testing may work. One big question we need to discuss: do we want to the unit tests to make live requests to our online platforms?

Advantages: allows us to test the platforms as well, and to make sure the APIs/webpages haven't broken or changed
Disadvantages: unit tests will run more slowly, since they need to actually download from the internet, our tests will break even if we did nothing wrong.

example.orchestration.yaml
```diff
@@ -2,6 +2,7 @@ steps:
   # only 1 feeder allowed
   feeder: gsheet_feeder # defaults to cli_feeder
   archivers: # order matters, uncomment to activate
+    - bluesky_archiver
```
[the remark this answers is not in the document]
msramalho: this was intentional as this is a no-setup archiver and would work out of the box. 
with a guideline/check in place this manual addition would not be needed indeed, but we can leave it for now.

src/auto_archiver/archivers/__init__.py
msramalho: good point, please open an issue for those changes.

msramalho:
I agree with merging this unit tests changes even if we refactor how those are setup etc, thanks for adding. 
I cannot approve my own PR so leaving that to @erinhmclark on this one.

src/auto_archiver/archivers/__init__.py
pjrobertson: Done, under 'backlog'

erinhmclark (approved):
Approved.

PR #161, opened 2024-12-23
adds better debug for wayback failures

msramalho:
I had this code written but it was not pushed before, followed hack into the wayback machine and consequent failed responses from their API.

### DIFF
--- a/src/auto_archiver/enrichers/wayback_enricher.py
+++ b/src/auto_archiver/enrichers/wayback_enricher.py
@@ -1,3 +1,4 @@
+import json
 from loguru import logger
 import time, requests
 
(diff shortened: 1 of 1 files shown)


src/auto_archiver/enrichers/wayback_enricher.py
```diff
@@ -70,11 +71,16 @@ def enrich(self, to_enrich: Metadata) -> bool:
...
-        job_id = r.json().get('job_id')
-        if not job_id:
-            logger.error(f"Wayback failed with {r.json()}")
+        try:
+            job_id = r.json().get('job_id')
+            if not job_id:
+                logger.error(f"Wayback failed with {r.json()}")
+                return False
+        except json.decoder.JSONDecodeError as e:
+            logger.error(f"Expected a JSON with job_id from Wayback and got {r.text}")
```
erinhmclark, line 80: Is it worth adding a length restriction to the text or streaming in case of large responses, or does it stay within a predictable length?

erinhmclark (approved):
That makes sense to me. I haven’t encountered this error in testing yet, but enriching the outputs seems like a good plan!

src/auto_archiver/enrichers/wayback_enricher.py
msramalho: predictable AFAIK.

PR #162, opened 2024-12-31
Fix two small issues

pjrobertson:
These are some small issues I came across whilst setting up/testing the project. I will add to this PR if I find any more. For bigger issues/features, I'll create new issues.

### DIFF
--- a/src/auto_archiver/archivers/twitter_archiver.py
+++ b/src/auto_archiver/archivers/twitter_archiver.py
@@ -108,9 +108,11 @@ class TwitterArchiver(Archiver):
         tweet = tie._extract_status(tweet_id)
         result = Metadata()
         try:
+            if not tweet.get("user") or not tweet.get("created_at"):
+                raise ValueError(f"Error retreiving post with id {tweet_id}. Are you sure it exists?")
             timestamp = datetime.strptime(tweet["created_at"], "%a %b %d %H:%M:%S %z %Y")
-        except Exception as ex:
-            logger.warning(f"Failed to get timestamp: {type(ex).__name__} occurred. args: {ex.args}")
+        except (ValueError, KeyError) as ex:
+            logger.warning(f"Unable to parse tweet: {str(ex)}\nRetreived tweet data: {tweet}")
             return False
                 
         result\
--- a/src/auto_archiver/archivers/twitter_archiver.py
+++ b/src/auto_archiver/archivers/twitter_archiver.py
@@ -147,10 +147,16 @@ class TwitterArchiver(Archiver):
         tie = TwitterIE(downloader)
         tweet = tie._extract_status(tweet_id)
         result = Metadata()
+        try:
+            timestamp = datetime.strptime(tweet["created_at"], "%a %b %d %H:%M:%S %z %Y")
+        except Exception as ex:
+            logger.warning(f"Failed to get timestamp: {type(ex).__name__} occurred. args: {ex.args}")
+            return False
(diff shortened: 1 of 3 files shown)


src/auto_archiver/formatters/templates/macros.html
```diff
@@ -3,7 +3,7 @@
 {% for url in m.urls %}
 {% if url | length == 0 %}
 No URL available for {{ m.key }}.
-{% elif 'http' in url %}
+{% elif 'http://' in url or 'https://' in url or url.startswith('/') %}
```
msramalho, line 6: good catch!

src/auto_archiver/archivers/twitter_archiver.py
```diff
@@ -147,10 +147,16 @@ def download_yt_dlp(self, item: Metadata, url: str, tweet_id: str) -> Union[Meta
         tie = TwitterIE(downloader)
         tweet = tie._extract_status(tweet_id)
         result = Metadata()
+        try:
+            timestamp = datetime.strptime(tweet["created_at"], "%a %b %d %H:%M:%S %z %Y")
+        except Exception as ex:
+            logger.warning(f"Failed to get timestamp: {type(ex).__name__} occurred. args: {ex.args}")
+            return False
```
msramalho, line 154: assumption here that if the datetime parse fails the response was actually broken, is it so?

I see this is the previous scenario but with more granular logging.

pjrobertson: Yes, I haven't changed the functionality, just made clearer logging (I think I got an error on this at one point, which is why I changed it for clearer logging). Something for us to discuss going forwards. I'll flag this on the project so I can discuss with @erinhmclark - i.e.: logging and how to deal with potentially broken/incomplete/unexpected responses (is an 'unexpected' response better than nothing?)

pjrobertson: OK, so running on an invalid tweet gives the following response from youtubedlp

```
(Pdb) !tweet
{'user': {}, 'card': {}, 'quoted_status': {}, 'retweeted_status': {}}
```

We need some kind of better validation to make sure the tweet is valid. Right now it'll fail because the timestamp isn't parsed correctly (it's empty) but that's a bit brittle. Will add it to this PR

pjrobertson:
OK, improved logging and pushed another commit along with unit tests to clarify the edge case where parsing the timestamp was failing (it's actually because the tweet didn't exist). Good catch, thanks!

[WIP] tag removed. Should be good to go

pjrobertson:
Note: this now also includes a fix for the 'download_syndication' method, which was previously failing. Follows the discussion at https://github.com/JustAnotherArchivist/snscrape/issues/996#issuecomment-2211358215

tests/archivers/test_twitter_archiver.py
```diff
@@ -70,8 +69,42 @@ def test_youtube_dlp_archiver(self):
...
+        )
 
+    def test_download_nonexistend_tweet(self):
+        # this tweet does not exist
+        url = "https://x.com/Bellingcat/status/17197025860711058"
+        response = self.archiver.download(self.create_item(url))
+        self.assertFalse(response)
+
+    def test_download_malformed_tweetid(self):
+        # this tweet does not exist
```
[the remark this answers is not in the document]
pjrobertson: Yep, I'm switching over to pytest in the unittest PR, we can do that there

Issue #155, opened 2024-11-29
ImportError: cannot import name 'verify' from 'OpenSSL.crypto' (unknown location)


djhmateer:
After running a `pipenv update` yesterday I got:

`ImportError: cannot import name 'verify' from 'OpenSSL.crypto' (unknown location)` when starting the app.

I've found that adding a force version to the Pipenv file fixes it:

```bash
# force this version
# probably used by requests below
# otherwise get ImportError: cannot import name 'verify' from 'OpenSSL.crypto' (unknown location)
pyopenssl = "==24.2.1"

requests = {extras = ["socks"], version = "*"}
```

https://pypi.org/project/pyOpenSSL/ - verify has been dropped in the latest version.

pjrobertson:
Confirmed that `pip install pyOpenSSL==24.2.1` fixes this issue. pyOpenSSL version 24.3.0 fails

Note: macOS, python 3.11.

PR #163, opened 2024-12-31
CI Unit tests

pjrobertson:
This is a WIP branch, pushing so I can test whether the workflows run successfully or not.

Also includes the following:

* Forces pyopenssl to 24.2.1, fixing #155 
* Upgrades numpy, which was causing a dependency issue on install
* Pins youtube-dlp to a specific version, which was causing a dependency issue on install


pjrobertson:
This PR should be good to go. Note:

✅ Core tests (not download) are passing - they run on all compatible python versions (3.10-3.13)
❌ Download tests are failing (because sensitive content tests are failing - they require logging in to see) - they run only on lowest supported python version 3.10
✅ Switched to pytest, but not using VCR for request recording (it doesn't work well with recording youtube-dlp requests, although I did make a PR on [vcrpy](https://github.com/kevin1024/vcrpy/issues/902) about it
❌ I have not yet refactored the unit tests to tidy up URLs etc. I think that will be another commit, I am still trying to figure out the best structure for unit tests

Notes:
* Core tests are run on every PR and push to main
* Download tests are run once per week, on Monday at 2:35pm UTC (set to Monday to pick up any issues from the previous week)

erinhmclark (dismissed):
Looking good! There are a few things in the tests which could be refactored since moving to using PyTest, but we do that in a future PR and get the CI up and running first.

tests/databases/test_csv_db.py
```diff
@@ -0,0 +1,32 @@
...
+import unittest
+
+from auto_archiver.databases.csv_db import CSVDb
+from auto_archiver.core import Metadata
+
+
+
+class TestCSVdb(unittest.TestCase):
+
+    def setUp(self):
```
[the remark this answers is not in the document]
pjrobertson: Yeah, I've kind of been in two minds about completely getting rid of unittest or keeping it in. I saw that youtube-dlp still uses [unittest and the structure](https://github.com/yt-dlp/yt-dlp/blob/master/test/test_cache.py#L26), yet they run all the [tests with pytest](https://github.com/yt-dlp/yt-dlp/blob/master/devscripts/run_tests.py#L32) and I was wondering why. Just looked into it further, and I think for them it's just legacy code.

I have switch over fully in the latest commit

tests/archivers/test_twitter_archiver.py
```diff
@@ -1,64 +1,37 @@
...
 
 from .test_archiver_base import TestArchiverBase
 
-class TestTwitterArchiver(TestArchiverBase, unittest.TestCase):
+class TestTwitterArchiver(TestArchiverBase):
 
     archiver_class = TwitterArchiver
     config = {}
+    @pytest.mark.parametrize("url, expected", [
+        ("https://t.co/yl3oOJatFp", "https://www.bellingcat.com/category/resources/"),  # t.co URL
```
erinhmclark, line 13: We could even add the comments into the parameterisation for speedier debugging:

```
    @pytest.mark.parametrize(
        "input_url, expected_url, description",
        [
            (
                "https://t.co/yl3oOJatFp",
                "https://www.bellingcat.com/category/resources/",
                "Should expand t.co URLs",
            ),
            
            ...
            
            assert archiver.sanitize_url(input_url) == expected_url, description
```

tests/archivers/test_twitter_archiver.py
```diff
@@ -135,36 +109,32 @@ def test_download_tweet_no_media(self):
...
-
-        post = self.archiver.download(self.create_item(url))
+        post = self.archiver.download(make_item(url))
         self.assertValidResponseMetadata(
             post,
             "This month's Bellingchat Premium is with @KolinaKoltai. She reveals how she investigated a platform allowing users to create AI-generated child sexual abuse material and explains why it's crucial to investigate the people behind these services https://t.co/SfBUq0hSD0 https://t.co/rIHx0WlKp8",
             datetime.datetime(2024, 12, 24, 13, 44, 46, tzinfo=datetime.timezone.utc)
         )
 
+    @pytest.mark.xfail(reason="Currently failing, sensitive content requires logged in users/cookies - not yet implemented")
```
erinhmclark, line 121: Nice!

erinhmclark (approved):
Looks good!

tests/archivers/test_twitter_archiver.py
pjrobertson: You're right. Nice one. Missed this before I merged, I'll save it for later!

PR #164, opened 2024-12-31
Migrate to Poetry

erinhmclark:
Trialing migration to Poetry from Pipenv.

### Reasoning
- Fewer points of contact, containing packaging and scaffolding
- Allows more complex dependency management, generally better at resolving conflicts
- Custom dependency groups which will help when increasing tests

### Things to Check
- Currently as a minor release but might be major?
- Is this the best way to implement in docker (strict versioning)

### Next
It's currently just changing to poetry with minimal other changes, but future PRs should:
- Restriction versions for stability/ security
- Automated dependency bumps and security checks


pjrobertson:
I'm just looking at this in the context of reviewing dependencies (esp related to python version) and also to familiarise myself with poetry (not used it before). Couple of thoughts that came up:

1. Which python versions do we want to support? Right now it's set to ~3.10 (i.e. >3.10,<3.11), but the tool does actually work with 3.11. Do we want to use `>=3.10,<3.12`. I think the config file should also have `requires-python` to throw an error if users are on the wrong python version (see [here](https://python-poetry.org/docs/pyproject/#requires-python))
2. Which python version do we want to build on? I don't have an answer to this, just bringing it up - should we build on the lowest compatible version or highest?
3. Should we use pypi's newly introduced [trusted publishing](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) instead? There's a [GH action](https://github.com/marketplace/actions/pypi-publish#trusted-publishing) already for doing this
4. I was going through the [poetry docs](https://python-poetry.org/docs/dependency-specification/#projectdependencies-and-toolpoetrydependencies), and they suggest using `project.dependencies` instead of `tool.poetry.dependencies` for newer versions of poetry
5. `project.urls` seems to be missing in the config - the docs say that using `homepage`, `repository` etc. are deprecated, and `project.urls` should be used instead. See [example](https://python-poetry.org/docs/pyproject/#homepage)
6. We could also add the `project.scripts` field with `auto-archiver = 'auto_archiver.__main__:main` (not sure if that is right and what the [benefit of doing this is though](https://python-poetry.org/docs/pyproject/#scripts)

I'm still looking into the `snscrape` issue that is blocking python >3.12 compatibility, so hopefully in the near future we don't have to restrict to `<3.12`

erinhmclark:
> I'm just looking at this in the context of reviewing dependencies (esp related to python version) and also to familiarise myself with poetry (not used it before). Couple of thoughts that came up:
> 
> 1. Which python versions do we want to support? Right now it's set to ~3.10 (i.e. >3.10,<3.11), but the tool does actually work with 3.11. Do we want to use `>=3.10,<3.12`. I think the config file should also have `requires-python` to throw an error if users are on the wrong python version (see [here](https://python-poetry.org/docs/pyproject/#requires-python))
> 2. Which python version do we want to build on? I don't have an answer to this, just bringing it up - should we build on the lowest compatible version or highest?
> 3. Should we use pypi's newly introduced [trusted publishing](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) instead? There's a [GH action](https://github.com/marketplace/actions/pypi-publish#trusted-publishing) already for doing this
> 4. I was going through the [poetry docs](https://python-poetry.org/docs/dependency-specification/#projectdependencies-and-toolpoetrydependencies), and they suggest using `project.dependencies` instead of `tool.poetry.dependencies` for newer versions of poetry
> 5. `project.urls` seems to be missing in the config - the docs say that using `homepage`, `repository` etc. are deprecated, and `project.urls` should be used instead. See [example](https://python-poetry.org/docs/pyproject/#homepage)
> 6. We could also add the `project.scripts` field with `auto-archiver = 'auto_archiver.__main__:main` (not sure if that is right and what the [benefit of doing this is though](https://python-poetry.org/docs/pyproject/#scripts)
> 
> I'm still looking into the `snscrape` issue that is blocking python >3.12 compatibility, so hopefully in the near future we don't have to restrict to `<3.12`

@pjrobertson Thanks for the review.
So it seems that poetry have just released their 2.0.0 version (on 2025-01-05) with the new project section!
With the interest of modernising and as it isn't merged in yet, it makes sense to try to get that updated in this PR.

That sounds good to move to trusted publishing in future -  I think we need to update the publisher configurations on PyPi if it's not granted already.

msramalho:
1. the more backwards compatible we can be the better but I think `>=3.10` is a good starting point
2. I think building with the lowest supported version is best for compatibility, but I haven't had bad experiences with that tbh. though CI testing should run on all supported ones
3. yup, I can updated as needed (see screenshot below for the things needed)

image

pjrobertson:
Just installed poetry and got this working. Very smooth :)

I also installed with `pip install -e .` in a virtualenv (which I like, since I can run auto-archiver directly) and that was working well :)

I'm yet to check the docker installation, but will try to do so early next week.

pjrobertson:
Just saw: the `README` still refers to pipenv under the `### Option 3 - local installation` section. Not sure if you're planning on updating that here, or in the documentation PR

erinhmclark:
@pjrobertson That's great to hear the poetry installation was working smoothly for you!

I pushed some changes to the Dockerfile earlier today which I think are alright for this PR, but I also opened an issue for any future optimisation for it. 

I was leaving the README updates for the documentation PR, but maybe it makes sense to remove the Pipenv references in this PR for consistency.

I'm still leaving this hanging as a WIP, but just need to setup the new PyPi publishing and confirm versioning is done correctly - the docker-publish workflow should pull the version from a git tag in the release but want to double check that doesn't risk breaking any existing implementations!

pjrobertson (approved):
✅ Docker image works
✅ wheel built with poetry works

Looks good! Approved :)

PR #165, opened 2025-01-07
Remove snscrape from the twitter_archiver

pjrobertson:
This removes snscrape from the twitter archiver

    1. snscrape twitter downloader no longer works (ref: https://github.com/JustAnotherArchivist/snscrape/issues/1045)
    2. snscrape limits python to versions <3.12

Note: this only removes snscraper from the code, and not from the project. As we are switching to poetry, I have not touched the pipenv files so as to avoid merge confliccts. @erinhmclark if you agree, then you can remove snscrape from the new poetry config files in your PR :)

Note: this PR also contains:

1. Some small updates to the test classes for achivers (starting to create a base class). Note this does not yet separate code tests and response validation. That is WIP
2. Fixes a potential bug whereby tweets archived by the twitter_archiver, using ytdlp aren't marked as 'success'(fully) archived, causing the orchestrator to continue to check other archivers for valid archives of the tweet. I'm not sure if this is intentional behavior or not. We should probably only allow a fixed type of value in 'Metadata.status' (e.g. either `None` or `archiver_name: success` to avoid edge case errors.

### DIFF
--- a/src/auto_archiver/archivers/twitter_archiver.py
+++ b/src/auto_archiver/archivers/twitter_archiver.py
@@ -2,7 +2,6 @@ import re, requests, mimetypes, json
 from typing import Union
 from datetime import datetime
 from loguru import logger
-from snscrape.modules.twitter import TwitterTweetScraper, Video, Gif, Photo
 from yt_dlp import YoutubeDL
 from yt_dlp.extractor.twitter import TwitterIE
 from slugify import slugify
@@ -49,7 +48,7 @@ class TwitterArchiver(Archiver):
         username, tweet_id = self.get_username_tweet_id(url)
         if not username: return False
 
-        strategies = [self.download_yt_dlp, self.download_snscrape, self.download_syndication]
+        strategies = [self.download_yt_dlp, self.download_syndication]
         for strategy in strategies:
             logger.debug(f"Trying {strategy.__name__} for {url=}")
             try:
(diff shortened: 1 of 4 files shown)


tests/archivers/test_archiver_base.py
```diff
@@ -0,0 +1,22 @@
...
+    def create_item(self, url, **kwargs):
+        item = Metadata().set_url(url)
+        for key, value in kwargs.items():
+            item.set(key, value)
+        return item
+    
+    def assertValidResponseMetadata(self, test_response, title, timestamp):
+        self.assertTrue(test_response.is_success())
+        self.assertEqual(title, test_response.get_title())
+        self.assertTrue(timestamp, test_response.get("timestamp"))
```
msramalho, line 22: we may need to updated this based if some archivers don't provide these, but let's keep it for now and adapt as needed later.

tests/archivers/test_archiver_base.py
```diff
@@ -0,0 +1,22 @@
+from auto_archiver.core import Metadata
+
+class TestArchiverBase(object):
```
msramalho, line 3: good abstraction for testing here!

tests/archivers/test_archiver_base.py
pjrobertson: Noted, yes - I though this could be a good 'template' and we can adjust as we add more tests

src/auto_archiver/archivers/twitter_archiver.py
```diff
@@ -153,8 +113,7 @@ def download_yt_dlp(self, item: Metadata, url: str, tweet_id: str) -> Union[Meta
             .set_timestamp(datetime.strptime(tweet["created_at"], "%a %b %d %H:%M:%S %z %Y"))
         if not tweet.get("entities", {}).get("media"):
             logger.debug('No media found, archiving tweet text only')
-            result.status = "twitter-ytdl"
-            return result
+            return result.success("twitter-ytdl")
```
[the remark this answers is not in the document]
pjrobertson: Removed this commit (no longer exists)

pjrobertson:
Commit that altered the 'status' for twitter-ytdl has been removed. I think this PR should be good to go.

@erinhmclark can you please review, and also confirm if you're good to remove the snscrape dependency in your poetry PR directly? Thanks!

Note: unit tests are still WIP. Once this is merged into main, I will rebase my feat/unittests branch, and then improve the unit tests. I have a working solution to use a record/reply technqiue for dealing with API responses.

pjrobertson:
P.S. I will add additional twitter_archiver tests (for sensitive content etc. as discussed) later, in the unit tests PR. I want to confirm the structure for storing live responses and record/replay on the unit tests first :)

tests/archivers/test_twitter_archiver.py
```diff
@@ -0,0 +1,95 @@
...
+
+from .test_archiver_base import TestArchiverBase
+
+
+class TestTwitterArchiver(TestArchiverBase, unittest.TestCase):
+
+    archiver_class = TwitterArchiver
+    config = {}
+    
+    def test_sanitize_url(self):
```
erinhmclark, line 14: In future we could move the test URLs out to a dictionary / csv and iterate over them to make it a bit more readable, and make it easier to add new test cases.

erinhmclark (approved):
@pjrobertson Absolutely that makes sense, I will remove the snscrape requirement from the poetry environment in the other PR.

Also sounds like a very good idea restrict `Metadata.status` to values.

tests/archivers/test_twitter_archiver.py
erinhmclark: I know you addressed these are a WIP in the description, so can keep the conversation for the next PR!

pjrobertson: Good idea, we should have somewhere with all our test urls, as opposed to littering the code with them! Will keep this in mind for the next PR :)

pjrobertson:
Thanks Erin for the review :) I've merged this into main. I will now rebase and work on the CI stuff more deeply!

PR #172, opened 2025-01-15
Add docker-compose for easy building and running of docker image in dev

pjrobertson:
Just use `docker compose up` to build and run. It'll only regenerate an image if


erinhmclark (approved):
Tested locally and looking good :)

PR #175, opened 2025-01-15
Create generic archiver for all valid youtube-dl URLs, add truthsocial extractor, unit tests for twitter_api extractor, utility methods for cleaning HTML and traversing objects

pjrobertson:
This is a proof of concept, but should hopefully do away with having to have separate archivers for things that ytdlp already does (e.g. tiktok, bluesky).

Currently: it fails for posts that don't have videos (e.g. bluesky, twitter) because youtubedlp is specifically for videos.

I put out a [PR on ytdlp](https://github.com/yt-dlp/yt-dlp/pull/12098#issuecomment-2593547516) just now to separate the extract_status logic and download_logic in the ytdlp bluesky class, and got a very quick reply from a maintainer:

> This is not a supported use case. These are private methods, and compat is not guaranteed. In fact, compat is almost certainly guaranteed to be broken with extractor code

And the PR got marked as 'do not merge'. Something to discuss, but it seems like we can only do away with separate archivers for sites that only do video (youtube, tiktok...)


src/auto_archiver/archivers/archiver.py
```diff
@@ -59,11 +67,28 @@ def download_from_url(self, url: str, to_filename: str = None, verbose=True) ->
...
-        assert d.status_code == 200, f"got response code {d.status_code} for {url=}"
-        with open(to_filename, 'wb') as f:
-            f.write(d.content)
-        return to_filename
+        try:
+            d = requests.get(url, stream=True, headers=headers)
+            d.raise_for_status()
+
+            # Peek at the first 256 bytes
+            first_256 = d.raw.read(256)
```
msramalho, line 75: is this a standard size for extension guessing?

src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -33,7 +33,115 @@ def configs() -> dict:
...
         }
+    
+    def download_additional_media(self, ie: str, video_data: dict, metadata: Metadata) -> Metadata:
+        """
+        Downloads additional media like images, comments, subtitles, etc.
 
+        Creates a 'media' object and attaches it to the metadata object.
+        """
+
+        # TODO: should we download all thumbnails, or just the chosen thumbnail?
```
msramalho, line 44: single cover is enough for now.

src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -33,7 +33,115 @@ def configs() -> dict:
...
+        """
+
+        base_keys = ['formats', 'thumbnail', 'display_id', 'epoch', 'requested_downloads',
+                     'duration_string', 'thumbnails', 'http_headers', 'webpage_url_basename', 'webpage_url_domain',
+                     'extractor', 'extractor_key', 'playlist', 'playlist_index', 'duration_string', 'protocol', 'requested_subtitles',
+                     'format_id', 'acodec', 'vcodec', 'ext', 'epoch', '_has_drm', 'filesize', 'audio_ext', 'video_ext', 'vbr', 'abr',
+                     'resolution', 'dynamic_range', 'aspect_ratio', 'cookies', 'format', 'quality', 'preference', 'artists',
+                     'channel_id', 'subtitles', 'tbr', 'url', 'original_url', 'automatic_captions', 'playable_in_embed', 'live_status',
+                     '_format_sort_fields', 'chapters', 'uploader_id', 'uploader_url', 'requested_formats', 'format_note',
+                     'audio_channels', 'asr', 'fps', 'was_live', 'is_live', 'heatmap', 'age_limit', 'stretched_ratio']
```
msramalho, line 70: from these I'd keep any that could describe the who/when so, that I see, that's `uploader_id`, `uploader_url` if that's not saved elsewhere and is in fact ids/usernames/public links.

src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -33,7 +33,115 @@ def configs() -> dict:
...
+        if ie == 'TikTok':
+            return base_keys + []
```
msramalho, line 72: this makes sense, may be something we lazily maintain but that's okay, it should capture the biggest chunk from the start

src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -33,7 +33,115 @@ def configs() -> dict:
...
+        """
+        Creates a Metadata object from the give video_data
+        """
+
+        # first add the media
+        result = self.download_additional_media(ie, video_data, result)
+
+        # keep the full title, no need for the shortened title (?)
+        video_data['title'] = video_data.pop('fulltitle', video_data.get('title'))
+        result.set_title(video_data.pop('title', url))
```
msramalho, line 86:
```suggestion
        result.set_title(video_data.pop('title', ""))
```
no need to duplicate URL info here.

src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -33,7 +33,115 @@ def configs() -> dict:
...
+        # extract comments if enabled
+        if self.comments:
+            result.set("comments", [{
+                "text": c["text"],
+                "author": c["author"], 
+                "timestamp": datetime.datetime.fromtimestamp(c.get("timestamp"), tz = datetime.timezone.utc)
+            } for c in video_data.get("comments", [])])
+
+        # then add the common metadata
+        if (timestamp := video_data.pop("timestamp", None)):
```
msramalho, line 105:
```suggestion
        if timestamp := video_data.pop("timestamp", None):
```

src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -33,7 +33,115 @@ def configs() -> dict:
...
+            timestamp = datetime.datetime.fromtimestamp(timestamp, tz = datetime.timezone.utc).isoformat()
+            result.set_timestamp(timestamp)
+        if (upload_date := video_data.pop("upload_date", None)):
```
msramalho, line 108:
```suggestion
        if upload_date := video_data.pop("upload_date", None):
```

src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -33,7 +33,115 @@ def configs() -> dict:
...
+    
+    def add_metadata(self, ie: str, video_data: dict, url:str, result: Metadata) -> Metadata:
+        """
+        Creates a Metadata object from the give video_data
+        """
+
+        # first add the media
+        result = self.download_additional_media(ie, video_data, result)
+
+        # keep the full title, no need for the shortened title (?)
```
msramalho, line 84: are we sure no info is lost here? thinking other platforms too

src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -70,17 +178,16 @@ def download(self, item: Metadata) -> Metadata:
...
         info = ydl.extract_info(url, download=True)
-
         if "entries" in info:
             entries = info.get("entries", [])
             if not len(entries):
                 logger.warning('YoutubeDLArchiver could not find any video')
                 return False
         else: entries = [info]
 
+        ie = info['extractor_key']
```
msramalho, line 188: let's rename this variable to something more explicit referring to extractor

src/auto_archiver/archivers/youtubedl_archiver.py
```diff
@@ -104,22 +211,11 @@ def download(self, item: Metadata) -> Metadata:
...
-            #TODO: fix deprecated timestamp, 
-            timestamp = datetime.datetime.fromtimestamp(timestamp, tz = datetime.timezone.utc).isoformat()
-            result.set_timestamp(timestamp)
-        if (upload_date := info.get("upload_date")):
-            upload_date = datetime.datetime.strptime(upload_date, '%Y%m%d').replace(tzinfo=datetime.timezone.utc)
-            result.set("upload_date", upload_date)
+        result = self.add_metadata(ie, info, url, result)
+        extractor_name = "yt-dlp"
+        if ie:
+            extractor_name += f"--{ie}IE"
```
msramalho, line 217: as this is mostly user-facing we should shorten/simplify as much as possible, 
```suggestion
            extractor_name += f"_{ie}"
```

msramalho (changes requested):
looking good, minor changes. 
are you adding the platform specific changes to this PR? (tiktok...)

pjrobertson:
> are you adding the platform specific changes to this PR? (tiktok...)

Yes, that was the thinking. If the general direction looks good to you, I'll also include the tiktok one (it's actually already here/working) and perhaps the Twitter one (?)

src/auto_archiver/archivers/archiver.py
pjrobertson: This was a typo, it should be 261, all that's [required to detect](https://github.com/h2non/filetype.py?tab=readme-ov-file#features). But yes, that's all that's needed to guess filetype. No point loading more than that (and wasting memory) to determine the filetype!

src/auto_archiver/archivers/youtubedl_archiver.py
pjrobertson: OK, I've removed the comment 👍

src/auto_archiver/archivers/youtubedl_archiver.py
pjrobertson: Tbh I'm not sure, it's worth probably keeping both - you're right.

pjrobertson:
OK, I've changed those items you mentioned, and have also started refactoring the code (quite a lot of changes). In summary:

1. Renamed `youtubedl_archiver` to `base_archiver` - happy to modify as you see fit
2. Moved the new `base_archiver` into its own folder (module) to kind of show how modules might look in the future (note that this is just to show what it might look like, but the code all still works and orchestrator remains untouched for now)
3. Removed the tiktok_archiver file - the `base_archiver` can do it all
4. Removed the twitter_archiver file - the `base_archiver` can do it all
5. Removed the twitter_api_archiver file - the `base_archiver` can do it all
6. Removed the bluesky_archiver - the `base_archiver` can do it all (once my upstream [here](https://github.com/yt-dlp/yt-dlp/pull/12098) is merged)

I think we're moving in the right direction, basically using youtubedlp as the main archiver, and only requiring additional archivers for where youtubedlp doesn't work.

In the future, we may also be able to remove:
1. vk archiver
2. Telegram archiver
3. facebook
4. Instagram

But I haven't touched those for now

src/auto_archiver/archivers/archiver.py
msramalho: interesting!
I haven't tested but did you check if mimetypes can achieve the same thing with file chunks? seeing this method https://docs.python.org/3/library/mimetypes.html#mimetypes.MimeTypes.readfp may be worth quickly testing to avoid adding a new dependency

src/auto_archiver/archivers/base_archiver/bluesky.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    Iterates over image(s) or video in a Bluesky post and downloads them        
+    """
+    media = []
+    embed = post.get("record", {}).get("embed", {})
+    image_medias = embed.get("images", []) + embed.get("media", {}).get("images", [])
+    video_medias = [e for e in [embed.get("video"), embed.get("media", {}).get("video")] if e]
+
+    for image_media in image_medias:
+            image_media = _download_bsky_file_as_media(image_media["image"]["ref"]["$link"], post["author"]["did"])
```
msramalho, line 37: indent issue here?

src/auto_archiver/archivers/twitter_api_archiver.py
msramalho, line 1: this archiver, or at least the logic in it should be kept as it's a totally different approach to fetching the data, in this case using an official API

src/auto_archiver/archivers/base_archiver/base_archiver.py
```diff
@@ -0,0 +1,291 @@
...
+        Returns False for the GenericIE, which as labelled by yt-dlp: 'Generic downloader that works on some sites'
+        """
+        return any(self.suitable_extractors(url))
+
+    def create_metadata_for_post(self, info_extractor: InfoExtractor, video_data: dict, url: str) -> Metadata:
+        """
+        Standardizes the output of the ytdlp InfoExtractor to a common format
+        """
+        if info_extractor.ie_key() == 'Bluesky':
+            return bluesky.create_metadata(video_data, self, url)
```
msramalho, line 151: am I correct in assuming that when getting the metadata from the info_extractor for post (vs for video) that the unnecessary keys cleaned up above are not present, or any form of those?

src/auto_archiver/archivers/base_archiver/base_archiver.py
```diff
@@ -0,0 +1,291 @@
...
+            # twitter kwargs are tweet_id
+            twid = ie_instance._match_valid_url(url).group('id')
+            # TODO: if ytdlp PR https://github.com/yt-dlp/yt-dlp/pull/12098 is merged, change to _extract_post
+            post_data = ie_instance._extract_status(twid=twid)
+
+        elif info_extractor.ie_key() == 'TikTok':
+            pass
+
+        else:
+            # lame attempt at trying to get data for an unknown extractor
```
msramalho, line 177:
```suggestion
            # lame attempt at trying to get data for an unknown extractor
            # TODO: test some more video platforms and see if there's any improvement to be made
```
in case this has not been tested for others.

msramalho (changes requested):
main change is not removing the official twitter api logic, otherwise looks good and we can iterate improvements over this demo module

src/auto_archiver/archivers/youtubedl_archiver.py
pjrobertson: They're not stored anywhere else. I'll remove them from here

src/auto_archiver/archivers/base_archiver/base_archiver.py
pjrobertson: You are correct, the unecessary keys are only for the 'video_data' which is the format returned by ytdlp for valid videos.

I don't think there's any harm in trying to remove the keys that don't exist - a few extra CPU cycles. But the next step would be to extend the `keys_to_clean` method to remove keys from the bluesky `post` json (i.e. the API response) that we also don't want. I realise I haven't done that here.

pjrobertson: Actually this is what the bsky `post` API gets. They look fine - nothing to clean. Below is bsky and twitter - let me know if you think anything else needs removing from either of them?
Screen Region 2025-01-17 at 11 49 15
Screen Region 2025-01-17 at 11 51 42

src/auto_archiver/archivers/base_archiver/base_archiver.py
```diff
@@ -0,0 +1,296 @@
...
+    
+    def suitable_extractors(self, url: str) -> list[str]:
+        """
+        Returns a list of valid extractors for the given URL"""
+        for info_extractor in yt_dlp.YoutubeDL()._ies.values():
+            if info_extractor.suitable(url) and info_extractor.working():
+                yield info_extractor
+        
+    def suitable(self, url: str) -> bool:
+        """
```
erinhmclark, line 144: This looks good. I was thinking in future we could bring it up the chain of processing, or even add an optional codelist into the source for known inputs.

erinhmclark (commented):
It's looking good, it's great to get so much pulled out into one place.

Were you changing anything in the orchestration.yaml? Step wasn't picking up the youtubedl_archiver as a subclass, I think due to the
TODO: cannot find subclasses of child.subclasses
Not sure if the plan is to change to process 'base_archiver' in the orchestrator or something?

src/auto_archiver/archivers/base_archiver/base_archiver.py
```diff
@@ -0,0 +1,296 @@
...
+            logger.debug(f'ytdlp exception which is normal for example a facebook page with images only will cause a IndexError: list index out of range. Exception is: \n  {e}')
+            return False
+        
+        if result:
+            extractor_name = "yt-dlp"
+            if info_extractor:
+                extractor_name += f"_{info_extractor.ie_key()}"
+
+            if self.end_means_success:
+                result.success(extractor_name)
```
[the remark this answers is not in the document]
pjrobertson: There's actually a subtle difference between "extractor_name: success" and "extractor_name". Miguel explained it in the AA chat a few days back:

> this should have been better documented as I can't say 100%, to understand it would require testing with some real cases. 
> The plausible reason for not being .success("twitter-ytl") is that this method was not working as expected to return the media and this would defer the logic to a "retrying" of fetching the video with the ytdlp archiver. 

But I do think we should standardise things better. Using an unknown "{extractor_name}" format for things that *don't* succeed could likely cause confusion. I created a note in the project on this.

src/auto_archiver/archivers/base_archiver/base_archiver.py
```diff
@@ -0,0 +1,296 @@
...
+
+from loguru import logger
+
+from . import bluesky, twitter
+from auto_archiver.archivers.archiver import Archiver
+from ...core import Metadata, Media, ArchivingContext
+
+
+class BaseArchiver(Archiver):
+    name = "youtubedl_archiver" #left as is for backwards compat
```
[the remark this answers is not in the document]
pjrobertson: I like GenericArchiver. BaseArchiver isn't ideal, because it makes it sound like it's the base class (as you asked). I can reword this.

Also, I like the idea of switching to extractor instead of archiver - to be in line with ytdlp and also I think it's clearer. Let's discuss with Miguel :)

src/auto_archiver/archivers/base_archiver/base_archiver.py
```diff
@@ -0,0 +1,296 @@
...
+            try:
+                cover_image_path = self.download_from_url(thumbnail_url)
+                media = Media(cover_image_path)
+                metadata.add_media(media, id="cover")
+            except Exception as e:
+                logger.error(f"Error downloading cover image {thumbnail_url}: {e}")
+
+        return metadata
+
+    def keys_to_clean(self, extractor_key: str, video_data: dict) -> dict:
```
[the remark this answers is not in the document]
pjrobertson: Maybe in the future video_data will be needed (e.g. some archivers may want to know what's already in video_data before they can decide what to clean). Don't want to too aggressively remove it just yet I don't think!

src/auto_archiver/archivers/base_archiver/base_archiver.py
```diff
@@ -0,0 +1,296 @@
...
+        # first add the media
+        result = self.download_additional_media(extractor_key, video_data, result)
+
+        # keep both 'title' and 'fulltitle', but prefer 'title', falling back to 'fulltitle' if it doesn't exist
+        result.set_title(video_data.pop('title', video_data.pop('fulltitle', "")))
+
+        # then add the platform specific additional metadata
+        for key, mapping in self.video_data_metadata_mapping(extractor_key, video_data).items():
+            if isinstance(mapping, str):
+                result.set(key, eval(f"video_data{mapping}"))
```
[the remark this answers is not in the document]
pjrobertson: Yeah I thought about this, but the `mapping` being used is hard-coded in the code itself. It's not taking user input, so I think it's ok. Some of the fields are double nested, like `['post']['media']` and I thought this was the easiest way to do it. But maybe there's a better way?

src/auto_archiver/archivers/base_archiver/base_archiver.py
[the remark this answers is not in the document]
erinhmclark: Oh the renaming to extractor was accidental from me in this context! But in general I do prefer extractors as a name for these processors, especially as the whole process is considered an archiver it's good to clearly separate the terminology.

src/auto_archiver/archivers/generic_archiver/generic_archiver.py
```diff
@@ -0,0 +1,297 @@
...
+        return any(self.suitable_extractors(url))
+
+    def create_metadata_for_post(self, info_extractor: InfoExtractor, video_data: dict, url: str) -> Metadata:
+        """
+        Standardizes the output of the ytdlp InfoExtractor to a common format
+        """
+        if info_extractor.ie_key() == 'Bluesky':
+            return bluesky.create_metadata(video_data, self, url)
+        if info_extractor.ie_key() == 'Twitter':
+            return twitter.create_metadata(video_data, self, url)
```
msramalho, line 158: tiktok missing here?

msramalho (approved):
happy with how it's looking and okay to merge if you are not planning to add any more features here (eg the tiktok comment I left).

src/auto_archiver/archivers/generic_archiver/generic_archiver.py
pjrobertson: This is a good question. So actually it's not needed for tiktok, since all tiktok pages have valid videos, and the 'standard' video_data is returned by the ytdl tiktok extractor. So no further processing is required.

But you've made me realise I can abstract out and make the code clearer. I will do that

pjrobertson:
> happy with how it's looking and okay to merge if you are not planning to add any more features here (eg the tiktok comment I left).

Only question mark here is that my upstream ytdlp mergePR (needed for the bluesky changes here) hasn't been merged yet. I've just pushed them. Worst case is I copy over the single method with a TODO for now, since this is all changing fairly quickly.

msramalho:
use your own judgement over how long you expect the PR to take vs how much code you have to port here/additional work later, and also if this is blocking/creating headaches for the next tasks.

src/auto_archiver/archivers/generic_archiver/truth.py
```diff
@@ -0,0 +1,31 @@
+import datetime
+
+from auto_archiver.utils import clean_html, traverse_obj
+from auto_archiver.core.metadata import Metadata
+from auto_archiver.archivers.archiver import Archiver
+
+def create_metadata(post: dict, archiver: Archiver, url: str) -> Metadata:
+    """
+    Creates metaata from a truth social post
```
msramalho, line 9:
```suggestion
    Creates metadata from a truth social post
```

src/auto_archiver/archivers/generic_archiver/truth.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    result = Metadata()
+    result.set_url(url)
+    timestamp = post['created_at'] # format is 2022-12-29T19:51:18.161Z
+    result.set_timestamp(datetime.datetime.strptime(timestamp, "%Y-%m-%dT%H:%M:%S.%fZ"))
```
msramalho, line 20:
```suggestion
    result.set_timestamp(post['created_at'])
```
parse_dt handles this format.
```python
>>> from dateutil.parser import parse as parse_dt
>>> parse_dt("2022-12-29T19:51:18.161Z")
datetime.datetime(2022, 12, 29, 19, 51, 18, 161000, tzinfo=tzutc())
```

src/auto_archiver/archivers/generic_archiver/truth.py
```diff
@@ -0,0 +1,31 @@
...
+    result.set('description', post['content'])
+    result.set('author', post['account']['username'])
+
+    for key in ['replies_count', 'reblogs_count', 'favourites_count', ('account', 'followers_count'), ('account', 'following_count'), ('account', 'statuses_count'), ('account', 'display_name'), 'language', 'in_reply_to_account', 'replies_count']:
+        if isinstance(key, tuple):
+            store_key = u" ".join(key)
```
msramalho, line 26:
```suggestion
            store_key = " ".join(key)
```
unicode prefix is only needed for python2. py3 assumes unicode by default

tests/archivers/test_generic_archiver.py
```diff
@@ -0,0 +1,191 @@
...
+        assert result is not False
+    
+    @pytest.mark.download
+    def test_truthsocial_download_single_image(self, make_item):
+        item = make_item("https://truthsocial.com/@mariabartiromo/posts/113861116433335006")
+        result = self.archiver.download(item)
+        assert len(result.media) == 1
+        assert result is not False
+
+    @pytest.mark.skip("Currently failing, multiple images are not being downloaded - this is due to an issue with ytdlp extractor")
```
msramalho, line 128: is there an issue on their repo we can link to here?

msramalho (changes requested):
minor changes only.

msramalho (approved):

src/auto_archiver/archivers/generic_archiver/generic_archiver.py
```diff
@@ -0,0 +1,322 @@
...
+            if data.get('is_live', False) and not self.livestreams:
+                logger.warning("Livestream detected, skipping due to 'livestreams' configuration setting")
+                return False
+            # it's a valid video, that the youtubdedl can download out of the box
+            result = self.get_metadata_for_video(data, info_extractor, url, ydl)
+
+        except Exception as e:
+            logger.debug(f'Issue using "{info_extractor.IE_NAME}" extractor to download video (error: {repr(e)}), attempting to use extractor to get post data instead')
+            try:
+                result = self.get_metatdata_for_post(info_extractor, url, ydl)
```
msramalho, line 277:
```suggestion
                result = self.get_metadata_for_post(info_extractor, url, ydl)
```

src/auto_archiver/archivers/generic_archiver/generic_archiver.py
```diff
@@ -0,0 +1,322 @@
...
+            video_data.pop(clean_key, None)
+        
+        # then add the rest of the video data
+        for k, v in video_data.items():
+            if v:
+                result.set(k, v)
+
+        return result
+
+    def get_metatdata_for_post(self, info_extractor: Type[InfoExtractor], url: str, ydl: yt_dlp.YoutubeDL) -> Metadata:
```
msramalho, line 153:
```suggestion
    def get_metadata_for_post(self, info_extractor: Type[InfoExtractor], url: str, ydl: yt_dlp.YoutubeDL) -> Metadata:
```

src/auto_archiver/archivers/generic_archiver/generic_archiver.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        Calls into the ytdlp InfoExtract subclass to use the prive _extract_post method to get the post metadata.
+        """
+
+        ie_instance = info_extractor(downloader=ydl)
+        post_data = None
```
msramalho, line 159:
```suggestion
```

msramalho (changes requested):

msramalho (approved):

PR #177, opened 2025-01-16
Add Sphinx documentation and publish to RTD.

erinhmclark:
Creating a fresh PR  adding Sphinx documentation with RTD integration.

### Summary
- Adds Sphinx and autoapi
- Add docs requirements group to poetry environment. We could further separate some such as autobuild into a docs-dev (or just move to dev?)
- Currently contains a generated documentation file of configs. I haven't put this script into this PR as it's more a proof of concept and we're planning to restructure, but in theory this or the equivalent could be generated within the docs.

### Run locally 

`poetry install`

```
sphinx-build -b html docs/source docs/_build/html
```

Then either manually:
`open docs/_build/html/index.html`

Or run autobuild to automatically detect changes when modifying the docs:

```
poetry run sphinx-autobuild docs/source docs/_build/html
```


erinhmclark:
Once we get the .readthedocs.yaml into main we can watch future builds on RTD.
The current README.md is loaded in the root of the docs, but still needs some internal paths fixed

pjrobertson:
This looks great! Some minor comments (pretty minor - I think this can already be merged!)

1. I got these two folders after building, do they need ignoring/commiting?   `/docs/build/` and `/docs/source/autoapi/`
2. For me, the built files were put in `docs/build/html/index.html` not `build/html/index.html` as you state in the PR - is this an issue with my build settings or a typo?
4. Configs seems redundant in the side panel, it's the same page as Configuration, and it's repeated twice:
Screen Region 2025-01-16 at 17 50 36

5. I'd put User Guidelines at the top in the sidebar!
6. Is there a way to make `Api Guidelines` not show the `src` part in the path?
Screen Region 2025-01-16 at 17 52 10
7. How about a 'Building the Docs' section in the Readme? (p.s. the PR says use `sphinx-build -b html docs/source docs/build/html` but I had to add `poetry run` at the start of course

I was going to comment on some of the `config` docs, but I think since we will be changing this quite a lot, there's no need to change it right now.

erinhmclark:
@pjrobertson most of those things were me lazily copying my instructions from different setups - apologies!

I will update the original description to fix those, and indeed docs/_make and docs/source/autobuild should be in .gitignore (I will do that in a min) and add a section to the README for it.

The Config docs were created by me running a hacky local script as mostly they're in the return statements rather than set as properties so I don't know if they can be pulled into the documentation via autoapi! If they do stay in a similar implementation we can tidy up that script and add it into the doc builds, but I think let's leave the automation out until then. However if there's something incorrect/ misleading I can manually edit that document too!

pjrobertson:
Just one small comment on the 'Configs' section in the 'Configuration' file (can we also rename 'Configs' to something more intuitive)?

And then the removal of the `src` prefix from the autoapi info - I think this can be done by removing `__init__.py` from the src folder. It doesn't need to be there, and it looks like it is just a hangup from [project restructuring anyway](https://github.com/bellingcat/auto-archiver/commits/d1e4dde3f60c92aa796d239dfb9d0cdc55d3d0b3/src/__init__.py)

pjrobertson (approved):
Great stuff!

Issue #180, opened 2025-01-21
S3 upload fail - MissingContentLength when calling the PutObject operation


djhmateer:
https://github.com/boto/boto3/issues/4398

I've temporarily downgraded boto3 to 1.35.99 to get around all S3 file uploads failing.

msramalho:
thanks @djhmateer we're pushing several breaking changes atm including converting from pipenv to poetry and we're using [1.35.99 there](https://github.com/bellingcat/auto-archiver/blob/113a4db25108a194a37366e2b952e8af3badbc74/poetry.lock#L154-L163). in the meantime, thanks for opening the issue as it can help others quickly identify the solution.

we should do a new release with all the new changes documented and tested in a few weeks, so in the meantime this issue applies to https://github.com/bellingcat/auto-archiver/releases/tag/v0.12.0 and thanks for the patience there.

msramalho:
I did not get this error running yday's version  https://pypi.org/project/boto3/1.36.3/ tested on both DigitalOcean and S3

djhmateer:
Ah yes it is working now for me too. Thanks @msramalho - much appreciated for the update

PR #181, opened 2025-01-21
Add ubuntu-latest to the matrix of test runners

pjrobertson:
tests on ubuntu-latest are currently failing, due to the issue in oscrypto: https://github.com/wbond/oscrypto/issues/78#issuecomment-1756317472

Not sure if we want to merge this in, or wait for a solution before we merge

### DIFF
--- a/.github/workflows/tests-core.yaml
+++ b/.github/workflows/tests-core.yaml
@@ -11,10 +11,12 @@ on:
 
 jobs:
   tests:
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         python-version: ["3.10", "3.11", "3.12"]
+        os: [ubuntu-22.04, ubuntu-latest]
     defaults:
       run:
         working-directory: ./
--- a/poetry.lock
+++ b/poetry.lock
@@ -152,34 +152,34 @@ lxml = ["lxml"]
 
 [[package]]
 name = "boto3"
-version = "1.35.99"
+version = "1.36.3"
 description = "The AWS SDK for Python"
 optional = false
 python-versions = ">=3.8"
 groups = ["main"]
 files = [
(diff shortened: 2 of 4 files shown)


msramalho:
both https://github.com/wbond/certvalidator
and https://github.com/wbond/oscrypto
seem unmaintained.

the source dependency issue is in certvalidator that installs oscrypto (where the issue is)

alledgedly, this commit (not published to pypi) https://github.com/wbond/oscrypto/commit/d5f3437ed24257895ae1edd9e503cfb352e635a8 solves the issue (not personally verified)

can we fix the oscrypto version to that one even if it is installed as a dependency? that would be a quick fix

---

alternative is to find another way to achieve the same as certvalidator is currently doing, that may be a more involved task so I'd put behind the major modularization work we are doing at the moment.

msramalho:
also I'd not merge this PR until we can make sure the test suite will be working

msramalho (approved):

pjrobertson:
Sorry, I missed this yesterday. Changes look good, but is this perhaps an opportunity to do away with oscrypto (certvalidator). You said this also fixes the boto3 issue? I checked boto3 and oscrypto isn't a dependency on it.

In my ['remove dependencies' PR](https://github.com/bellingcat/auto-archiver/pull/169#issue-2786805961) I mentioned that the timestamping_enricher has another old module `asn1crypto`. Perhaps a TODO is to rewrite timestamping_enricher to remove `asn1crypto` and `certvalidator` (and subsequently `oscrypto`). Thoughts? Shall we open an issue on it?

msramalho:
Let's do any refactoring of timestamping_enricher in a dedicated branch, and yes worth an issue but with this PR the current codebase will work on ubunutu 24 which was a blocker. 

the boto3 issue is not connected to oscrypt but doing a poetry install auto-bumped that version to one that works see #180

PR #182, opened 2025-01-23
modifies base docker image to use browsertrix 1.4.2

msramalho:
- Had to switch to poetry-venv in dockerfile for it to build. 
- `blockAds` seems to interfere with cloudflare bot prevention, so I'm removing it for now

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM webrecorder/browsertrix-crawler:1.0.4 AS base
+FROM webrecorder/browsertrix-crawler:1.4.2 AS base
 
 ENV RUNNING_IN_DOCKER=1 \
     LANG=C.UTF-8 \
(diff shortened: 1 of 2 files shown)


erinhmclark (approved):
That's great to get the base image bumped up to the latest!

Strange about it only building with venv though.  
I think with creating the venv there might be further optimisations we can make, but I can investigate that another time.

pjrobertson:
The official docs recommend using `venv` (see https://python-poetry.org/docs/#ci-recommendations ) so I think this is probably the right way.

Dockerfile
```diff
@@ -29,21 +29,23 @@ ENV POETRY_NO_INTERACTION=1 \
     POETRY_VIRTUALENVS_CREATE=1
 
 
-RUN pip install --upgrade pip && \
-    pip install "poetry>=2.0.0,<3.0.0"
+# Create a virtual environment for poetry and install it
+RUN python3 -m venv /poetry-venv && \
```
pjrobertson, line 33: you could abstract out the `/poetry-venv` repetivie part by doing what the docs say and:
```
export POETRY_HOME=/opt/poetry
python3 -m venv $POETRY_HOME
$POETRY_HOME/bin/pip install --upgrade pip
$POETRY_HOME/bin/pip install "poetry>=2.0.0,<3.0.0"
```
```

PR #183, opened 2025-01-23
More mainifests

pjrobertson:
Erin's work, I'm just creating the PR so we can discuss :)


src/auto_archiver/base_modules/storage.py
```diff
@@ -4,40 +4,17 @@
...
 class Storage(Step):
     name = "storage"
-    PATH_GENERATOR_OPTIONS = ["flat", "url", "random"]
-    FILENAME_GENERATOR_CHOICES = ["random", "static"]
-
-    def __init__(self, config: dict) -> None:
-        # without this STEP.__init__ is not called
-        super().__init__(config)
-        assert self.path_generator in Storage.PATH_GENERATOR_OPTIONS, f"path_generator must be one of {Storage.PATH_GENERATOR_OPTIONS}"
-        assert self.filename_generator in Storage.FILENAME_GENERATOR_CHOICES, f"filename_generator must be one of {Storage.FILENAME_GENERATOR_CHOICES}"
```
pjrobertson, line 25: This is awesome. We can remove so much boilerplate code now!  Just need `choices = ` in the manifest configs, and it'll do all the checks for us!

src/auto_archiver/modules/generic_extractor/twitter.py
```diff
@@ -6,7 +6,7 @@
 
 from auto_archiver.core.metadata import Metadata, Media
 from auto_archiver.utils import UrlUtil
-from auto_archiver.archivers.archiver import Archiver
+from auto_archiver.base_modules.extractor import Extractor
```
pjrobertson, line 9: How about just `auto_archiver.base.extractor` ? Since these aren't actually modules per se, they are just mixins. In fact, you don't even *have* to subclass from them anymore if you don't want to. They are just there for convenience and typing.

src/auto_archiver/modules/gsheet_processor/__manifest__.py
```diff
@@ -0,0 +1,58 @@
...
+    "type": ["feeder"],
+    "requires_setup": True,
+    "external_dependencies": {
+        "python": ["loguru", "gspread", "python-slugify"],
+    },
+    "configs": {
+        "allow_worksheets": {
+            "default": set(),
+            "help": "(CSV) only worksheets whose name is included in allow are included (overrides worksheet_block), leave empty so all are allowed",
+            "cli_set": lambda cli_val, cur_val: set(cli_val.split(","))
```
pjrobertson, line 12: these `cli_set` vals can be removed, they're not needed anymore. We should use the `type=XXX` key/value on these dicts now, since that's what argparse uses.

src/auto_archiver/base_modules/enricher.py
```diff
@@ -0,0 +1,31 @@
...
+
+Enrichers are optional but highly useful for making the archived data more powerful.
+"""
+from __future__ import annotations
+from dataclasses import dataclass
+from abc import abstractmethod, ABC
+from auto_archiver.core import Metadata, Step
+
+@dataclass
+class Enricher(Step, ABC):
```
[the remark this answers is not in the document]
erinhmclark: Ah so keeping it as an ABC would ensure that any subclasses enforce the implementation of any abstractmethods we define - I don't think it was actually needed here before as it would have inherited the property from the Step class, but without Step it might make sense to bring it back in!

src/auto_archiver/modules/api_db/__manifest__.py
```diff
@@ -16,7 +16,7 @@
             "allow_rearchive": {"default": True, "help": "if False then the API database will be queried prior to any archiving operations and stop if the link has already been archived"},
             "store_results": {"default": True, "help": "when set, will send the results to the API database."},
             "tags": {"default": [], "help": "what tags to add to the archived URL",
-                     "type": lambda val: set(val.split(",")),
+                     "type": "auto_archiver.utils.parse_csv_to_set",
```
pjrobertson, line 19: we don't need this anymore. Since we're using argparse, when options are passed on the command line, argparse does this all for us.

PR #185, opened 2025-01-29
Refactor auto-archiver to use a modular structure for feeders/extractors/enrichers etc.

### LINKED ISSUE #130: improve config parsing to accommodate type casting and validation rules
As it stands the logic to handle the `configs()` method of each Step (like Archivers, Enrichers, ...) does not cast int/float/bool and uses only strings. Likewise it's proven quite useful to specify validation rules in the configs rather than relying on the init method of each Step to do so. 

This issue is on adding both of those configuration parsing mechanisms and refactoring existing Steps that are doing internal casting/validation.

### LINKED ISSUE #156: No hashing algorithm defined for use in HtmlFormatter when HashEnricher not included

### LINKED ISSUE #173: Database.failed() method called without all required params
On this line the databases are receiving a failed call without all the required params `(item, reason)`: 
https://github.com/bellingcat/auto-archiver/blob/05e0c9de935b2f9ea598263324124e73a8ec1775/src/auto_archiver/core/orchestrator.py#L67

leading to an unhandled exception that breaks normal execution.

Take this as an optional param or enforce it in another way.

pjrobertson:
Also:

* View a full list of options and config settings by running `auto-archiver --help` 
* Quicker startup of auto-archiver, by only installing the enabled modules
* New/additional modules can be enabled/disabled on the fly using command line arguments such as  `--extractor=new_temporary_extractor`
* You can now run auto-archiver first time without needing to create an orchestration.yaml file first, auto-archiver will set up a basic one so you can get started in no time 🚀
* You can choose to update your config to file/rewrite your latest command line args to it (so you don't have to keep passing them) by setting the `-s` or `--store` flag on the command line
* All config settings are checked/validated at startup, along with module dependencies, to make sure you have setup modules correctly. Developers: you can create your own 'validators' to check values, and have the program check against them. Just drop them into the `validators.py` file (uses argparse [type under the hood](https://docs.python.org/3/library/argparse.html#type)). Fixes #130 
* Improved errors/warnings if there are any config/dependency issues and more helpful guidance when getting started (plus: colour output)
*You can now log directly to files using the `logging: file` option. Set the logging level as well using `logging: level`
* Fixed up unit tests + added a few new ones
* Set your own custom module folder with `module_paths=/my/own/modules/` to allow you to easily extend auto archiver with new modules. Simply create a new module, place it in that folder then pass the folder path on the command line/save it in your orchestration.yaml
* Update yt-dlp to latest version and factor out obsolete code from the `bluesky.py` dropin (no longer needed, as we can call the `yt-dlp` extractor directly
* Create an 'authentication' mechanism for storing site authentication information (username/password, api keys, cookies etc.) in the config file, which can then be used by all extractors. Set your authentication details in one place, and all modules will be able to use them without having to set multiple additional configs!
* Fixes using the screenshot enricher with facebook (click the 'accept cookies' and close the 'login' modals to take a proper screenshot). Screenshot enricher can now also use your cookies to login to sites to take screenshots (cookies from a cookiejar file or directly from your browser).
* New CSV feeder, which allows you to input CSV files of URLs (more info [here](https://github.com/bellingcat/auto-archiver/blob/load_modules/src/auto_archiver/modules/csv_feeder/__manifest__.py))

Developers:

* Removes `ArchivingContext` completely, and now passes contextual information around using the metadata `metadata.set_context('key', 'val')` or `metadata.get_context('key', 'val')`
* Easily get authentication information in your module using `self.auth_for_site(url)` - this returns a dict with the different authentication options available depending on what the user has set (e.g. username/password, cookiejar, api_key, cookie).
* Add your own modules without touching the auto-archiver project structure. Pass `--module_paths /my/modules/folder/` on the command line, or add `module_paths: /my/modules/folder/` to your config file.
* Easily leverage the power of yt-dlp by using it to extract information for a website without having to create an entirely new extractor. Subclass the 'GenericDropin from the `generic_extractor` module, and add the two methods `extract_post` and `create_metadata`, which both have access to yt-dlp's 'InfoExtractor' under the hood. An example of how to do this can be found [here](https://github.com/bellingcat/auto-archiver/blob/main/src/auto_archiver/archivers/generic_archiver/bluesky.py)
* Allow modules to have 'dependencies' of either a) other modules b) 3rd party python packages c) binary packages, by settings them in the 'dependencies' field in the __manifest__.py. (Note: fixes #156 )

Other Fixes:
* Fix a bug sending items to the database (Fixes #173 )

### DIFF
--- a/src/auto_archiver/modules/screenshot_enricher/__manifest__.py
+++ b/src/auto_archiver/modules/screenshot_enricher/__manifest__.py
@@ -4,7 +4,7 @@
     "requires_setup": True,
     "dependencies": {
         "python": ["loguru", "selenium"],
-        "bin": ["chromedriver"]
+        "bin": ["geckodriver"]
     },
     "configs": {
             "width": {"default": 1280, "help": "width of the screenshots"},


src/auto_archiver/modules/gsheet_feeder/gsheet_feeder.py
```diff
@@ -68,7 +68,7 @@ def __iter__(self) -> Metadata:
                     folder = os.path.join(folder, slugify(self.sheet), slugify(wks.title))
 
                 m.set_context('folder', folder)
-                m.set_context('worksheet', {"row": row, "worksheet": gw})
+                m.set_context('gsheet', {"row": row, "worksheet": gw})
                 yield m
```
erinhmclark, line 72: I had made some changes (and this fix) over in #189 - sorry it was probably unnecessary to branch off this branch!

erinhmclark (approved):

PR #190, opened 2025-02-10
Docs improvement

pjrobertson:
Improvement to docs based on new orchestrator structure, switch to using md instead of rst, better structure and linking between files


erinhmclark (approved):
Looking good! I think we can get it merged in and iterate on any additions we want to make.

erinhmclark (approved):

PR #196, opened 2025-02-15
makes orchestrator.run return the results to allow for code integration

msramalho:


msramalho:
@pjrobertson merging with some orchestrator tests failing, please have a look at what broke, the only addition was to the return value of the .run function. AP

pjrobertson:
I think we need to figure out a better way to make this work for command line usage + code integration. We want the command line usage to throw errors and give user feedback, but we don't want that for code integration.

With these changes now, you won't get *any* feedback on the command line. Try running:

`auto-archiver BLAHBLAH --invalid-url 99999` 

You will see nothing on the command line. In short - code integration shouldn't be using the `run()` method. We should have another entry point. Will look into this more next week, but just so you know - this change will mean command line interaction doesn't work right now

Issue #200, opened 2025-02-15
`ValueError: There is no existing handler with id 0` when executed >1 times programmatically


msramalho:
`ArchivingOrchestrator.run()` is not idempotent, so sequential programmatic calls raise exceptions. 

Known case:

 https://github.com/bellingcat/auto-archiver/blob/6d43bc7d4daef9a1baf52b0399085250a85bef0c/src/auto_archiver/core/orchestrator.py#L239

will throw `ValueError: There is no existing handler with id 0`

if the auto-archiver is executed > 1 times in the same programatic setup.

pjrobertson:
Fixed by #210

PR #210, opened 2025-02-19
Fix issue #200

pjrobertson:
Small PR, just wraps a try/catch around the logging, and doesn't re-add a logger if it's already been added. 

Note: this should no longer be needed since code-implementations should only call `orchestrator.setup()` **once**, but this is a good check to have just in case. I've added in a warning to remind people if they call setup() multiple times:

```python
if self.setup_finished:
     logger.warning("The `setup_config()` function should only ever be run once. \
                    If you need to re-run the setup, please re-instantiate a new instance of the orchestrator. \
                    For code implementatations, you should call .setup_config() once then you may call .feed() \
                    multiple times to archive multiple URLs.")
```

### DIFF
--- a/src/auto_archiver/core/orchestrator.py
+++ b/src/auto_archiver/core/orchestrator.py
@@ -95,6 +95,8 @@ class UniqueAppendAction(argparse.Action):
 
 class ArchivingOrchestrator:
 
+    setup_finished: bool = False
+    logger_id: int = None
     feeders: List[Type[Feeder]]
     extractors: List[Type[Extractor]]
     enrichers: List[Type[Enricher]]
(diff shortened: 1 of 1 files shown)


erinhmclark (approved):
Looks good!

pjrobertson:
I've just pushed another commit here. This does away with the ugly `_LAZY_LOADED_MODULES` variable, which was previously a global var, i.e. shared across all python instances. This meant that if you loaded a module with some config in one place:

```
o1 = ArchivingOrchestrator()
o1.setup(...)

# and then loaded a 2nd orchestrator
o2 = ArchivingOrchestrator()
o2.setup(...)

```
desplite the 2nd orchestrator appearing to setup correctly, it was actually using the modules already loaded (with configs set) from the `o1.setup()` step.

The changes here turn the `module.py` file into an actual class, and each ArchivingOrchestrator has its own 'ModuleFactor' to create modules and set configs just for itself. 🔓

erinhmclark (approved):
This is looking good!

PR #222, opened 2025-02-25
yt-dlp proposed extractor_args and PO Token client

erinhmclark:
## [PO Token client side plugin](https://github.com/coletdjnz/yt-dlp-get-pot)

This is only the client side plugin, it only notifies for client types when a token is needed, it does not create or provide a token [see here](https://github.com/coletdjnz/yt-dlp-get-pot?tab=readme-ov-file#installing)

Without this yt-dlp automatically would otherwise skip combinations which they know to require tokens.

## [Token provider](https://github.com/Brainicism/bgutil-ytdlp-pot-provider) 

To create and pass an actual token, you also need a token provider. For this we're using: bgutil-ytdlp-pot-provider
This token provider has two parts, one which passes the token to the client side plugin, yt-dlp-get-pot, and is installed as a package, then you also need the actual provider which is either run via a HTTP server which would need to run throughout the process (not ideal as it creates an extra requirement on the generic_extractor. This could be implemented as an optional module, where users can run the server separately however if we have the installed packages without the implementation this would lead to more errors and probably more blocking in cases that might otherwise have worked, or we only install the modules if this flag is turned on which can also lead to complications.)

### Token provider script

**Implementation**

The default implementation in Docker is getting the full bgutil-ytdlp-pot-provider git repo, and transpling the token generation script.

Locally / using pypi package the default is to not implement the update or creation of the token generation script, however the plugins (bgutil-ytdlp-pot-provider, which includes yt-dlp-get-pot) are still installed in the poetry environment.
This means that if you are running a version of the HTTP server, or have the pre-transpiled script set up see [here](https://github.com/Brainicism/bgutil-ytdlp-pot-provider?tab=readme-ov-file#a-http-server-option) in its default location or port, the bgutil-ytdlp-pot-provider plugin should be able to reach these without you needing to pass any additional args.

If you want the AA to pull the bgutil-ytdlp-pot-provider git repo, and ensure its the same version as the plugin version, you can pass an arg to the generic extractor to do the script method.

```
generic_extractor:
    bguils_po_token_method: "script"
```

### Auto Update of bgutil-ytdlp-pot-provider plugin

This PR also implements a check for updates for bgutil-ytdlp-pot-provider, using the same process and schedule as yt-dlp. This is to keep it up date with changes, and also to ensure that the version of the plugin installed in the environment matches the version of the server set up.


src/auto_archiver/modules/generic_extractor/generic_extractor.py
```diff
@@ -65,6 +70,21 @@ def update_ytdlp(self):
         except Exception as e:
             logger.error(f"Error updating yt-dlp: {e}")
 
+    def setup_token_script(self):
+        """Setup PO Token provider https://github.com/Brainicism/bgutil-ytdlp-pot-provider."""
+
+        if self.pot_provider == "bgutils":
+            # Check if the PO token generation script exists, set it up if not.
+            try:
+                subprocess.run(["bash", "scripts/potoken_provider/setup_pot_provider.sh"], check=True)
```
pjrobertson, line 79: This is cool! But as I was looking through this, I was thinking - that scripts folder doesn't exist if people install via pip/docker. You actually made me realise I made the same mistake on my [PR here,](https://github.com/bellingcat/auto-archiver/pull/255#issuecomment-2736358684) which I fixed by downloading directly from GH (this is how e.g. homebrew installs itself)

I'm wondering if the same should be done here - to download the script directly from GH. Alternatively, we could package the node stuff with the project - but then this won't work if the user doesn't have node/yarn installed (should we add that to the Docker file?)

src/auto_archiver/modules/generic_extractor/generic_extractor.py
```diff
@@ -65,6 +70,21 @@ def update_ytdlp(self):
...
+            except subprocess.CalledProcessError as e:
+                logger.error(f"Failed to setup PO Token script: {e}")
+                return
+
+            # Use the PO Token script in yt-dlp to fetch tokens on demand.
+            pot_script = os.path.join("scripts", "potoken_provider", "bgutil-provider", "build", "generate_once.js")
+            self.extractor_args.setdefault("youtube", {})["getpot_bgutil_script"] = pot_script
```
pjrobertson, line 86: instead of doing this, would it be possible just to set the default extractor_args in the __manifest__ to:

```
        "extractor_args": {
            "default": {
                 'youtube': {'getpot_bgutil_script' : pot_script},
             }
            "help": "Additional arguments to pass to the yt-dlp extractor. See https://github.com/yt-dlp/yt-dlp/blob/master/README.md#extractor-arguments.",
            "type": "json_loader",
        },
```

Or would that not work?

src/auto_archiver/modules/generic_extractor/generic_extractor.py
```diff
@@ -419,16 +442,20 @@ def download(self, item: Metadata) -> Metadata:
...
             "--write-auto-subs" if self.subtitles else "--no-write-auto-subs",
             "--live-from-start" if self.live_from_start else "--no-live-from-start",
-            "--proxy",
-            self.proxy if self.proxy else "",
-            f"--max-downloads {self.max_downloads}" if self.max_downloads != "inf" else "",
-            f"--playlist-end {self.max_downloads}" if self.max_downloads != "inf" else "",
         ]
 
+        # proxy handling
+        if self.proxy:
```
pjrobertson, line 448: nice, much easier to read than my `if else`

src/auto_archiver/modules/generic_extractor/generic_extractor.py
erinhmclark: I was expecting to add node/ yarn installation in Docker for this, but hadn't thought about the scripts not being packaged in pip/docker versions!
This script depends on the other having ran already (update_pot_provider.sh) or the bgutils-server files within scripts, that is if we wanted to add those to version control.
I was separating these two steps so that we weren't needing to clone the whole repo from gh within docker/ this setup process as that makes things tricky in it's own way.

src/auto_archiver/modules/generic_extractor/generic_extractor.py
```diff
@@ -444,6 +471,16 @@ def download(self, item: Metadata) -> Metadata:
...
 
+        # Applying user-defined extractor_args
+        if self.extractor_args:
+            for key, args in self.extractor_args.items():
+                logger.debug(f"Setting extractor_args: {key}")
+                if isinstance(args, dict):
+                    arg_str = ";".join(f"{k}={v}" for k, v in args.items())
+                else:
+                    arg_str = str(args)
+                ydl_options.extend(["--extractor-args", f"{key}:{arg_str}"])
```
pjrobertson, line 482: If I'm understanding this right, it adds an additional `--extractor-args` for each extractor in the dict. I'm assuming this works, i.e. having something like this is OK?

```
yt-dlp [options] --extractor-args facebook:a=b;c=d --extractor-args youtube:x=y;z=1
```

src/auto_archiver/modules/generic_extractor/generic_extractor.py
erinhmclark: So we can do, but then if the script isn't setup bgutils plugin will try to fetch a token each time and raise an error.
It depends if we can get this fully baked in, or whether we want it as optional.

src/auto_archiver/modules/generic_extractor/generic_extractor.py
erinhmclark: I was going by this https://github.com/yt-dlp/yt-dlp/blob/master/README.md#extractor-arguments
but maybe that was meant to be two examples and they are meant to be space separated?
e.g.
```
--extractor-args "youtube:player-client=tv,mweb;formats=incomplete twitter:api=syndication"
```

I think I tested it, but I will double check!

pjrobertson:
Just as a note/aside, I've just discovered that browsertrix already has and uses `node` v20 - so that is already installed in [the Docker image](https://github.com/webrecorder/browsertrix-browser-base/blob/main/Dockerfile#L23), as is yarn. That means that for this to work in docker, you just need to do add this to the dockerfile:

```
git clone --single-branch --branch 0.8.0 https://github.com/Brainicism/bgutil-ytdlp-pot-provider.git
cd bgutil-ytdlp-pot-provider/server/
yarn install --frozen-lockfile
npx tsc
```

:D

scripts/potoken_provider/setup_pot_provider.sh
```diff
@@ -0,0 +1,33 @@
...
+# Ensure the server directory exists
+if [ ! -d "$TARGET_DIR" ]; then
+    echo "Error: PO Token provider server directory is missing! Please run scripts/update_pot_provider.sh first."
+    exit 1
+fi
+
+cd "$TARGET_DIR" || exit 1
+
+# Check if dependencies need installation
+if [ ! -d "node_modules" ]; then
```
Brainicism, line 17: You should do a `yarn install` regardless of `node_module`'s existence. Sometimes Youtube will break things for us, and we need to update our main dependency (youtubei.js) to fix it. 

If no dependencies have changed, the `yarn install` is a no-op, so you won't be wasting any time on this regardless

scripts/potoken_provider/setup_pot_provider.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo "Installing dependencies..."
+    yarn install --frozen-lockfile
+else
+    echo "Dependencies already installed. Skipping yarn install."
+fi
+
+# Check if build directory exists and if transpiling is needed
+if [ ! -d "build" ]; then
```
Brainicism, line 25: How does this check if the build is up to date? Doesn't this only check if it has been transpiled once some indeterminate time ago?

Brainicism:
Sorry about the unsolicited review, but left a few comments on things that might have caused some pain down the road.

scripts/potoken_provider/setup_pot_provider.sh
Brainicism: Curious why you don't just merge these two bash scripts together. If the `update` script also runs the dependency install, and the transpilation, you don't need these special checks to tell if something has changed

scripts/potoken_provider/update_pot_provider.sh
```diff
@@ -0,0 +1,26 @@
...
+# Copy only the contents inside /server/ into bgutil-provider,
+# as this is the part containing the PO Token generation script
+echo "Copy /server/ contents into $TARGET_DIR..."
+cp -r "$BGUTIL_TEMP_DIR/server/"* "$TARGET_DIR/"
+
+# Clean up: remove the cloned repository as we only needed the /server/ contents
+echo "Cleaning up temporary files..."
+rm -rf "$BGUTIL_TEMP_DIR"
+
+echo "PO Token provider script is ready in: $TARGET_DIR/build"
```
Brainicism, line 25: Is this log misleading? `/build` wouldn't be available until after the `tsc`

erinhmclark:
> Sorry about the unsolicited review, but left a few comments on things that might have caused some pain down the road. Or I could be misunderstanding something and feel free to ignore everything I said 🙂

Hi @Brainicism !

Thank you so much for the review, it’s really great to get your input on this 😁

The reason for the overcomplicated setup here was to avoid cloning the full repo from our Docker image - though we’ve come to the conclusion that this might be the best approach!
 The ‘modules’ in our tool (e.g. here the generic_extractor which uses yt-dlp) are optional and set by passing a config, so we would ideally only run the server/ transpile the script when the generic_extractor is selected. We also have a Python package for the Auto Archiver so we were looking for an elegant way to get this working for both.
We want the Auto Archiver tool to require minimum setup for some modules, and the `generic_extractor` module otherwise works out of the box.

If we pull the full repo in our Dockerfile (native HTTP approach or transpiling in there) that should work nicely, then locally/ via the packaged version of the Auto Archiver users could optionally run one of the token provider methods :)

It’d be amazing to get your advice on the behaviour when the plugin installed but a server/ generation script isn’t available?

I see that the plugin still detects when a PO token is needed (i.e. yt-dlp-get-pot identifies certain clients that typically require a token), and then bgutil-ytdlp-pot-provider attempts to fetch or generate one, but if a provider isn’t available the actual YouTube request is skipped, is that right? 
I’m not too worried about the overhead of the attempt itself, but just wanted to check that it wouldn’t increase the risk of getting rate-limited or blocked in cases where the token provider isn’t set up!

Brainicism:
>  The ‘modules’ in our tool (e.g. here the generic_extractor which uses yt-dlp) are optional and set by passing a config, so we would ideally only run the server/ transpile the script when the generic_extractor is selected.

This makes sense. This only weird part to me is how the `update` script and the `setup` script are used in different stages? It looks like the `update` script which pulls the repo is run manually. Then the `setup` script runs automatically when the generic extractor is called. Why not just `setup` (install dependencies & transpile) during the `update` (pull/update repo)? You should only ever need to `setup` once per `update`. Combining these will decrease the time for the generic extractor to run too, since it won't have to spend 10-20s setting up, and you won't need that `self.setup_token_script()` in code anymore.  If the user did an `update`, they probably intend to `setup` anyway. 

> If we pull the full repo in our Dockerfile (native HTTP approach or transpiling in there) that should work nicely, then locally/ via the packaged version of the Auto Archiver users could optionally run one of the token provider methods :)

This should work fine!

>It’d be amazing to get your advice on the behaviour when the plugin installed but a server/ generation script isn’t available?
I see that the plugin still detects when a PO token is needed (i.e. yt-dlp-get-pot identifies certain clients that typically require a token), and then bgutil-ytdlp-pot-provider attempts to fetch or generate one, but if a provider isn’t available the actual YouTube request is skipped, is that right?

If there's no provider available, the request **will still go through**. It doesn't make sense to make lack of provider a blocker for yt-dlp, because not all formats require a PO token. The provider setup should be a one-time thing though.. you might just have to ask your users to double check that [it was set-up correctly](https://github.com/Brainicism/bgutil-ytdlp-pot-provider?tab=readme-ov-file#verification). It might be helpful to set-up both methods too, since they will fallback on each other if needed.

erinhmclark:
Hey @Brainicism !

>
> This makes sense. This only weird part to me is how the `update` script and the `setup` script are used in different stages? It looks like the `update` script which pulls the repo is run manually. Then the `setup` script runs automatically when the generic extractor is called. Why not just `setup` (install dependencies & transpile) during the `update` (pull/update repo)? You should only ever need to `setup` once per `update`. Combining these will decrease the time for the generic extractor to run too, since it won't have to spend 10-20s setting up, and you won't need that `self.setup_token_script()` in code anymore. If the user did an `update`, they probably intend to `setup` anyway.

Apologies, I was half way through a refactor before and had left it in a confusing state! 
It's now set up to pull the full repo and catch updates on a regular basis, as well as ensuring the plugin version installed matches the server version.

> 
> If there's no provider available, the request **will still go through**. It doesn't make sense to make lack of provider a blocker for yt-dlp, because not all formats require a PO token. The provider setup should be a one-time thing though.. you might just have to ask your users to double check that it was set-up correctly. It might be helpful to set-up both methods too, since they will fallback on each other if needed.

Okay great, I have added a note in our documentation to set up both the server and script methods where possible! 

Thanks again for all the advice on these! 😁

docs/source/how_to/authentication_how_to.md
```diff
@@ -106,5 +106,94 @@ Finally,Some important things to remember:
...
+
+**PyPi/ Local**:
+
+When using the Auto Archiver PyPI package, or running locally, you will need additional system requirements to run the token generation script, namely either Docker, or Node.js and Yarn.
+
+See the [bgutil-ytdlp-pot-provider](https://github.com/Brainicism/bgutil-ytdlp-pot-provider?tab=readme-ov-file#a-http-server-option) documentation for more details.
+
+⚠️WARNING⚠️: This will add the server scripts to the home directory of wherever this is running.
+
+- You can set the config option `"po_token_provider": true` under the `GenericExtractor` section of your config to "script" to enable the token generation script process locally.
```
pjrobertson, line 146: The wording here seems confusing. `true` isn't a valid value. Should it be `po_token_provider": default` ?

src/auto_archiver/modules/generic_extractor/generic_extractor.py
```diff
@@ -26,59 +29,138 @@ class GenericExtractor(Extractor):
...
+            )
+            return
+        try:
+            from importlib.metadata import version as get_version
+
+            plugin_version = get_version("bgutil-ytdlp-pot-provider")
+            base_dir = os.path.expanduser("~/bgutil-ytdlp-pot-provider")
+            server_dir = os.path.join(base_dir, "server")
+            version_file = os.path.join(server_dir, ".VERSION")
+            transpiled_script = os.path.join(server_dir, "build", "generate_once.js")
```
pjrobertson, line 121: looking at the bgutil docs again, it's possible to run the server with just

```
git clone --single-branch --branch 0.8.1 https://github.com/Brainicism/bgutil-ytdlp-pot-provider.git
cd bgutil-ytdlp-pot-provider/server/
yarn install --frozen-lockfile
npx tsc
node build/main.js
```

The dev over there seems to have added a warning about the generate_once.js script, so why don't we do away with the generate_once.js option, and when we're in docker we just run:

`node build/main.js`

Or alternatively, we do this in the docker image and then when the docker image is launched, this server is also set up? This implementation is fine, just thinking though it might be nice to only use the server method in both docker and outside of docker.

The docs could also be updated to remove saying to set `po_token_provider` to script, and it can just be left  as "run the server locally either with docker or with `node build/main.js` and run this script, it'll pick up the server automatically

docs/source/how_to/authentication_how_to.md
```diff
@@ -106,5 +106,94 @@ Finally,Some important things to remember:
...
+# Proof of Origin Tokens
+
+YouTube uses **Proof of Origin Tokens (POT)** as part of its bot detection system to verify that requests originate from valid clients. If a token is missing or invalid, some videos may return errors like "Sign in to confirm you're not a bot."
+
+yt-dlp provides [a detailed guide to POTs](https://github.com/yt-dlp/yt-dlp/wiki/PO-Token-Guide).
+
+### How we can add POTs to Auto Archiver
+This feature is enabled for the Generic Archiver via two yt-dlp plugins:
+
+- **Client-side plugin**: [yt-dlp-get-pot](https://github.com/coletdjnz/yt-dlp-get-pot)
```
pjrobertson, line 123: should we just add `yt-dlp-get-pot` to the pyproject requirements?

docs/source/how_to/authentication_how_to.md
```diff
@@ -106,5 +106,94 @@ Finally,Some important things to remember:
...
+{"poToken":"MlMxojNFhEJvUzGeHEkVRSK_luXtwcDnwSNIOgaUutqB7t99nmlNvtWgYayboopG6ZopZgmQ-6PJCWEMHv89MIiFGGlJRY25Fkwzxmia_8uYgf5AWf==","generatedAt":"2025-03-26T10:45:26.156Z","visitIdentifier":"ymCMy8OflKM"}
+[debug] [GetPOT] Fetching gvs PO Token for tv client
+```
+
+If it can't find the script or something, you'll see something like this:
+```shell
+[debug] [GetPOT] Fetching player PO Token for tv client
+WARNING: [GetPOT] BgUtilScript: Script path doesn't exist: /Users/you/bgutil-ytdlp-pot-provider/server/build/generate_once.js. Please make sure the script has been transpiled correctly.
+WARNING: [GetPOT] BgUtilHTTP: Error reaching GET http://127.0.0.1:4416/ping (caused by TransportError). Please make sure that the server is reachable at http://127.0.0.1:4416.
+[debug] [GetPOT] No player PO Token provider available for tv client
```
pjrobertson, line 194: I just tried running this with the node server running with `node build/main.js` and here are the logs I got. Not sure if this is an error, if if this is normal? The server is running, and I don't see a 'Error reaching GET' log. 

```
WARNING: [GetPOT] Failed to fetch player PO Token for tv client: No request handlers configured
WARNING: [GetPOT] Failed to fetch gvs PO Token for tv client: No request handlers configured
WARNING: [GetPOT] Failed to fetch player PO Token for ios client: No request handlers configured
WARNING: [GetPOT] Failed to fetch gvs PO Token for ios client: No request handlers configured
WARNING: [GetPOT] Failed to fetch player PO Token for web client: No request handlers configured
WARNING: [GetPOT] Failed to fetch gvs PO Token for web client: No request handlers configured
WARNING: [youtube+GetPOT] t6R5l7Y1O1w: nsig extraction failed: Some formats may be missing
         n = XX0qa9mSzeZysPtcD4 ; player = https://www.youtube.com/s/player/20830619/player_ias.vflset/en_US/base.js
```

src/auto_archiver/modules/generic_extractor/generic_extractor.py
erinhmclark: So users can run that or the docker server when running the python package, and the plugin will pick up on it without additional setup.
Within Docker I stuck with the generation script as otherwise it would depend on a persistant background process which might be hard to debug on failure/
This was what I mentioned on the call the other day, this works well for the current single threaded AA structure, but the server/ dind would be recommended for concurrency.

docs/source/how_to/authentication_how_to.md
erinhmclark: So bgutils installs this if it doesn't exist, but we can add it for clarity if you think that's best?

pjrobertson:
LGTM :)

pjrobertson (approved):

PR #224, opened 2025-02-25
Timestamping rewrite

pjrobertson:
This is a rewrite of the module to use `rfc3161-client` instead of `tsp_client`. Unit tests were written *prior* to the refactor to confirm that they still work afterwards.

Some notes:

1. This is waiting on two PRs to be merged into rfc3161-client: trailofbits/rfc3161-client#121 and trailofbits/rfc3161-client#120
2. This will not work with non-compliant TSAs, such as digicert. See https://github.com/trailofbits/rfc3161-client/issues/104#issuecomment-2621960840 for more info. 

Notes:

1. The issue we have is that pyopenssl dropped support for verifying signatures. Even the tsp_client author is complaining about this and [looking for a solution](https://github.com/pyca/cryptography/issues/12124)
3. `cryptography` is still quite a mess, the latest version (45) which is set to be released *soon* has a fix for it, that was implemented [here](https://github.com/pyca/cryptography/pull/11800/files)
4. rfc3161 might have some kind of solution in there.


src/auto_archiver/modules/timestamping_enricher/timestamping_enricher.py
```diff
@@ -20,6 +33,22 @@ class TimestampingEnricher(Enricher):
...
+        self.session.headers.update(
+            {
+                "Content-Type": "application/timestamp-query",
+                "User-Agent": f"Auto-Archiver {__version__}",
+                "Accept": "application/timestamp-reply",
+            }
+        )
+
+    def __del__(self) -> None:
+        """
```
erinhmclark, line 47: I think it's more reliable to specifically implement .cleanup() like with the extractor. Maybe we can make cleanup a BaseModule method so that there's a consistent cleanup option across modules?

src/auto_archiver/modules/timestamping_enricher/timestamping_enricher.py
```diff
@@ -70,30 +102,92 @@ def enrich(self, to_enrich: Metadata) -> None:
...
+        certificates = self.tst_certs(tsp_response)
 
-        validator = CertificateValidator(first_cert, intermediate_certs=intermediate_certs, validation_context=context)
-        path = validator.validate_usage({'digital_signature'}, extended_key_usage={'time_stamping'})
 
         cert_chain = []
-        for cert in path:
+        for cert in certificates:
             cert_fn = os.path.join(self.tmp_dir, f"{str(cert.serial_number)[:20]}.crt")
+            print(cert_fn)
```
erinhmclark, line 188: Should this be a log, or is it not needed?

src/auto_archiver/modules/timestamping_enricher/timestamping_enricher.py
```diff
@@ -70,30 +102,92 @@ def enrich(self, to_enrich: Metadata) -> None:
...
+            verifier = builder.build()
+
+            
+            try:
+                verifier.verify(timestamp_response, message_hash)
+                return certificate
+            except Rfc3161VerificationError as e:
+                logger.debug(f"Unable to verify Timestamp with CA {certificate.subject}: {e}")
+                continue
+
```
erinhmclark, line 150: Could remove the continue and return None - but maybe should it be raising an error if it reaches the return? I know it's caught in the `enrich` method, but might keep it more consistent to raise here then keep the return type as None

erinhmclark (approved):
I added a few comments, but looking good!

src/auto_archiver/modules/timestamping_enricher/timestamping_enricher.py
pjrobertson: You're right!

petertodd:
Just a small suggestion: if you do go ahead with adding OpenTimestamps support too, I'd suggest also re-timestamping RFC-3161 timestamps with OpenTimestamps. That'd allow them to be verified in the future even if a given RFC-3161 authority has their private key compromised.

msramalho:
is this PR still waiting for any dependency? or could use a review?

pjrobertson:
@erinhmclark  - I have bumped the RFC3161-client version to 1.0.1 and also bumped AA version to 0.13.9.

Can you please review, and if all good then we can merge and get a new release out!

erinhmclark (approved):
Approved!

erinhmclark (approved):

PR #226, opened 2025-02-25
Merge modules - GSheet, Atlos

erinhmclark:
Merging processes which share resources.
Currently this includes

## Gsheet Feeder Database
The gsheet feeder and database were previously two separate modules, but are now joined into one module.
You can still use the feeder without the database, and still need to use the feeder in order to use the database.

## Atlos 
Merge Atlos feeder, database and storage into one module.


pjrobertson:
>  Merge storage into this as well, or keep it separate?

Hmm... that's a tough one! I would probably say keep them separate, since maybe uploading to atlos would be something people want to do separately? Only thing is that then we'd still have duplicate configs of `api_token` and `atlos_url`

Also, what would we name it if it was 3-in-1?1

erinhmclark:
> > Merge storage into this as well, or keep it separate?
> 
> Hmm... that's a tough one! I would probably say keep them separate, since maybe uploading to atlos would be something people want to do separately? Only thing is that then we'd still have duplicate configs of `api_token` and `atlos_url`
> 
> Also, what would we name it if it was 3-in-1?1

So I was planning to merge just the feeder + database, and keep the storage separate, but looking at the structure it might make more sense for them to go together. 
Atlos groups the information around each media event, and so it seems the current atlos_storage and atlos_db implementation currently depend on the fed url, or the atlos id of that url `item.metadata.get("atlos_id")`, being available.

msramalho:
I support merging the three atlos modules into one, the coupled id just highlights that need

src/auto_archiver/modules/atlos_feeder_db_storage/atlos_feeder_db_storage.py
```diff
@@ -0,0 +1,147 @@
...
+                f"/api/v2/source_material/upload/{atlos_id}",
+                params={"title": media.properties},
+                files={"file": (os.path.basename(media.filename), file_obj)},
+            )
+        logger.info(f"Uploaded {media.filename} to Atlos with ID {atlos_id} and title {media.key}")
+        return True
+
+    def uploadf(self, file: IO[bytes], key: str, **kwargs: dict) -> bool:
+        """Upload a file-like object; not implemented."""
+        raise NotImplementedError("uploadf method is not implemented yet.")
```
pjrobertson, line 146: Should this just pass and not raise? Like it was before: https://github.com/bellingcat/auto-archiver/pull/226/files#diff-e48355e6600e1760e053262fc370a79c26290e6e2f0756ea1c3d93d69ca786fbL66

Otherwise this will fail, right?

src/auto_archiver/modules/atlos_feeder_db_storage/atlos_feeder_db_storage.py
```diff
@@ -0,0 +1,147 @@
...
+            json={
+                "metadata": {
+                    "processed": True,
+                    "status": "success",
+                    "results": self._process_metadata(item),
+                }
+            },
+        )
+        logger.info(f"Stored success for {item.get_url()} (ID {atlos_id}) on Atlos")
+
```
pjrobertson, line 106: Suggested to add a comment to separate the feeder/db/storage sections for readability? Like:


```
# ! Atlos Module - Storage

erinhmclark: Yeah that's a good plan! I have added that for the three sections

src/auto_archiver/modules/atlos_feeder_db_storage/atlos_feeder_db_storage.py
erinhmclark: Indeed! done.

pjrobertson (approved):
Looks good! I've only tested the gsheets module, but that works well - works nicely decoupled! (don't have atlos set up). 

I also snuck in a few commits to update the docs to note the module name changes, and one tiny commit to add an additional cookie button name for clicking ;-)

Issue #232, opened 2025-03-03
Screenshot Enricher Broken in Docker image


pjrobertson:
1. Set up docker with `docker pull bellingcat/auto-archiver`
2. Run docker with `docker run --rm -v .:/app/secrets bellingcat/auto-archiver --config secrets/orchestration.yaml https://twitter.com/bellingcat/status/1674700676612386816 --logging.level=DEBUG --extractors=generic_extractor --enrichers=screenshot_enricher`
3. Following error:

```
2025-03-03 14:27:35.273 | ERROR    | auto_archiver.core.orchestrator:archive:546 - ERROR enricher screenshot_enricher: Message: Unable to obtain driver for firefox; For documentation on this error, please visit: https://www.selenium.dev/documentation/webdriver/troubleshooting/errors/driver_location
: Traceback (most recent call last):
  File "/app/.venv/lib/python3.12/site-packages/selenium/webdriver/common/driver_finder.py", line 67, in _binary_paths
    output = SeleniumManager().binary_paths(self._to_args())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/selenium/webdriver/common/selenium_manager.py", line 47, in binary_paths
    args = [str(self._get_binary())] + args
                ^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/selenium/webdriver/common/selenium_manager.py", line 94, in _get_binary
    raise WebDriverException(f"Unsupported platform/architecture combination: {sys.platform}/{arch}")
selenium.common.exceptions.WebDriverException: Message: Unsupported platform/architecture combination: linux/aarch64


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/auto_archiver/core/orchestrator.py", line 544, in archive
    try: e.enrich(result)
         ^^^^^^^^^^^^^^^^
  File "/app/auto_archiver/modules/screenshot_enricher/screenshot_enricher.py", line 27, in enrich
    with self.webdriver_factory(
  File "/app/auto_archiver/utils/webdriver.py", line 103, in __enter__
    self.driver = CookieSettingDriver(cookies=self.auth.get('cookies'), cookiejar=self.auth.get('cookies_jar'),
[...]

PR #233, opened 2025-03-03
Docker webdriver aarch64

pjrobertson:
This downloads the correct version of geckodriver depending on the Docker arch, fixing #232 

This was only an issue for people running on Mac with M1 chips.

Personally, I still vote for doing away with multiple archs for easier maintenance. Sure, maybe for Mac users (and linux ARM users - are there many?) it'll mean a slightly slower run, but 🤷‍♂️

If we remove ARM builds, it'll also a) speed up the docker build and b) mean we can remove the geckodriver update dance (selenium supports geckodriver for linux x86 out of the box)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN add-apt-repository ppa:mozillateam/ppa && \
 	apt-get install -y --no-install-recommends firefox-esr && \
     ln -s /usr/bin/firefox-esr /usr/bin/firefox
 
-ARG GECKODRIVER_VERSION=0.35.0
+ARG GECKODRIVER_VERSION=0.36.0
 
 RUN if [ $(uname -m) = "aarch64" ]; then \
         GECKODRIVER_ARCH=linux-aarch64; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,16 +7,26 @@ ENV RUNNING_IN_DOCKER=1 \
     PYTHONFAULTHANDLER=1 \
     PATH="/root/.local/bin:$PATH"
 
+
+ARG TARGETARCH
+
 # Installing system dependencies
 RUN add-apt-repository ppa:mozillateam/ppa && \
 	apt-get update && \
     apt-get install -y --no-install-recommends gcc ffmpeg fonts-noto exiftool && \
 	apt-get install -y --no-install-recommends firefox-esr && \
-    ln -s /usr/bin/firefox-esr /usr/bin/firefox && \
-    wget https://github.com/mozilla/geckodriver/releases/download/v0.35.0/geckodriver-v0.35.0-linux64.tar.gz && \
+    ln -s /usr/bin/firefox-esr /usr/bin/firefox
+
+ARG GECKODRIVER_VERSION=0.35.0
+
+RUN if [ $(uname -m) = "aarch64" ]; then \
+        GECKODRIVER_ARCH=linux-aarch64; \
+    else \
+        GECKODRIVER_ARCH=linux64; \
+    fi && \
+    wget https://github.com/mozilla/geckodriver/releases/download/v${GECKODRIVER_VERSION}/geckodriver-v${GECKODRIVER_VERSION}-${GECKODRIVER_ARCH}.tar.gz && \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -7,16 +7,26 @@ ENV RUNNING_IN_DOCKER=1 \
     PYTHONFAULTHANDLER=1 \
     PATH="/root/.local/bin:$PATH"
 
+
+ARG TARGETARCH
```
erinhmclark, line 11: Do we want to add an option to override with an arg using this?

Dockerfile
```diff
@@ -7,16 +7,26 @@ ENV RUNNING_IN_DOCKER=1 \
...
+
+RUN if [ $(uname -m) = "aarch64" ]; then \
+        GECKODRIVER_ARCH=linux-aarch64; \
+    else \
+        GECKODRIVER_ARCH=linux64; \
+    fi && \
+    wget https://github.com/mozilla/geckodriver/releases/download/v${GECKODRIVER_VERSION}/geckodriver-v${GECKODRIVER_VERSION}-${GECKODRIVER_ARCH}.tar.gz && \
     tar -xvzf geckodriver* -C /usr/local/bin && \
     chmod +x /usr/local/bin/geckodriver && \
-    rm geckodriver-v* && \
```
erinhmclark, line 19: Would we want to remove the geckodriver tar file after? It's probably insignificant compared to the rest of the size but we may as well make it clean as possible!

src/auto_archiver/utils/webdriver.py
```diff
@@ -64,14 +72,26 @@ def get(self, url: str):
...
             # try and click the 'close' button on the 'login' window to close it
             close_button = self.find_element(By.XPATH, "//div[@role='dialog']//div[@aria-label='Close']")
             if close_button:
                 close_button.click()
+        else:
 
+            # for all other sites, try and use some common button text to reject/accept cookies
+            for text in ["Refuse non-essential cookies", "Decline optional cookies", "Reject additional cookies", "Accept all cookies"]:
+                try:
+                    accept_button = self.find_element(By.XPATH, f"//*[contains(translate(text(), 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz'), '{text.lower()}')]")
```
erinhmclark, line 89: Nice! 

If we want to extend it in the future, we could create a lookup csv with "host, cookie text" to make it more targeted/ extendable for different sites.

We could also add a WebDriverWait to ensure the button is clickable or just visible before interacting, in case there's a lag, something like 
```
 WebDriverWait(self, 5).until(EC.element_to_be_clickable((By.XPATH, xpath))).click()
```

Though sometimes the humble time.sleep(1) is as good, as sometimes selenium doesnt deem it as clickable, often when its layered/ encompassing elements, but then when you try to click it does still work.

erinhmclark (dismissed):
This is looking great and very useful, I added some comments but otherwise looks good!

Dockerfile
pjrobertson: This setting comes from `--platform=` or `DOCKER_DEFAULT_PLATFORM`, so you set those in your command if you want to change the platform. I think that's best!

Dockerfile
pjrobertson: you're right, that was inadvertently removed!

src/auto_archiver/utils/webdriver.py
pjrobertson: Yeah, this was just a 'lame' attempt at trying to do this. Tbh I don't know if it's actually useful/used so just picked a few random strings from twitter, facebook and a few other sites!

Good spot on the wait. I've added it in

pjrobertson:
OK, a few more changes pushed here. I snuck in one extra change to the default screenshot height. I think it will help take more useful screenshots.

Before:
429ca43d6ee3458ca3ec26c7

After:
00553f10ac934f0e8fd0aee8

erinhmclark (approved):
Nice! That's looking good with the larger screenshots too. I think there are certain combinations of settings + user agents which can help avoid detection - maybe that's something that could be covered in a how-to guide at some point if it becomes relevant.

Issue #215, opened 2025-02-20
sensitive tiktok videos are not getting downloaded


msramalho:
ytdlp is not currently working for sensitive tiktok videos.

is this something we can address?

https://github.com/yt-dlp/yt-dlp/issues/8678

ping me for an example video.

![Image](https://github.com/user-attachments/assets/1ac91849-4ad8-412e-8c32-219b685e2955)

PR #237, opened 2025-03-07
Adds new extractor for tiktok via unofficial API

msramalho:
closes https://github.com/bellingcat/auto-archiver/issues/215

uses an unofficial API to capture videos that generic_extractor cannot capture without cookies or other authentication data.

### DIFF
--- a/src/auto_archiver/modules/tiktok_tikwm_extractor/tiktok_tikwm_extractor.py
+++ b/src/auto_archiver/modules/tiktok_tikwm_extractor/tiktok_tikwm_extractor.py
@@ -1,6 +1,8 @@
+import re
 import requests
 from loguru import logger
 from datetime import datetime, timezone
+from yt_dlp.extractor.tiktok import TikTokIE
 
 from auto_archiver.core import Extractor
 from auto_archiver.core import Metadata, Media
@@ -10,15 +12,15 @@ class TiktokTikwmExtractor(Extractor):
     """
     Extractor for TikTok that uses an unofficial API and can capture content that requires a login, like sensitive content.
     """
+    TIKWM_ENDPOINT = "https://www.tikwm.com/api/?url={url}"
 
     def download(self, item: Metadata) -> Metadata:
         url = item.get_url()
-
-        # detect URLs that we definitely cannot handle
-        if 'tiktok.com' not in item.netloc:
+        
+        if not re.match(TikTokIE._VALID_URL, url):
             return False
 
-        endpoint = f"https://www.tikwm.com/api/?url={url}"
+        endpoint = TiktokTikwmExtractor.TIKWM_ENDPOINT.format(url=url)
 
         r = requests.get(endpoint)
         if r.status_code != 200:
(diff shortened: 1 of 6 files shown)


src/auto_archiver/modules/tiktok_tikwm_extractor/tiktok_tikwm_extractor.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    def download(self, item: Metadata) -> Metadata:
+        url = item.get_url()
+
+        # detect URLs that we definitely cannot handle
+        if 'tiktok.com' not in item.netloc:
+            return False
+
+        endpoint = f"https://www.tikwm.com/api/?url={url}"
```
pjrobertson, line 21: Might be nice to declare this in the class with TIKWM_ENDPOINT for readability/maintainability (if they change the API URL)


```python

class TiktokTikwmExtractor(Extractor):
    """
    Extractor for TikTok that uses an unofficial API and can capture content that requires a login, like sensitive content.
    """

   TIKWM_ENDPOINT = 'https://www.tikwm.com/api/?url={url}'

   ...
   endpoint = self.TIKWM_ENDPOINT.format(url=url)
   ...
```

src/auto_archiver/modules/tiktok_tikwm_extractor/tiktok_tikwm_extractor.py
```diff
@@ -0,0 +1,73 @@
...
+class TiktokTikwmExtractor(Extractor):
+    """
+    Extractor for TikTok that uses an unofficial API and can capture content that requires a login, like sensitive content.
+    """
+
+    def download(self, item: Metadata) -> Metadata:
+        url = item.get_url()
+
+        # detect URLs that we definitely cannot handle
+        if 'tiktok.com' not in item.netloc:
```
pjrobertson, line 18: Probably fine, but for sites that yt-dlp also processes, I've been checking their code for valid URLs. E.g. 

https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/extractor/tiktok.py#L623

Would mean you could use:

```
_VALID_URL.match(url)
```

Just an FYI, I think that's overkill for this. Worst case is a URL like "https://example.com/something/tiktok.com" would pass

pjrobertson:
LGTM! Nice unit tests :D

tests/extractors/test_tiktok_tikwm_extractor.py
```diff
@@ -0,0 +1,155 @@
+from datetime import datetime, timezone
+import time
+from unittest.mock import patch
```
erinhmclark, line 3: We've added pytest-mock which can make patching a bit tidier, but it's fine to keep using this too!

msramalho:
all comments addressed, relevant points thanks!

pjrobertson (approved):
Approved! :)

erinhmclark (approved):

PR #243, opened 2025-03-10
Unit tests for storage types + fix storage too long issues for local storage

### LINKED ISSUE #228: Local storage "filename too long" error.
When using local_storage if the website URL is too long it fails to store with a "filename too long" error.

Error seen on Mac but should be transversal. 
Also occurs when either of the path_generator options.

pjrobertson:
Fixes #228 

More importantly, I also found a bug in the base `Storage` class that meant the storage settings (like path_generator and file_generator) weren't being honoured.

I've fixed that + added unit tests.

It turns out, this 'filename too long' bug could appear for *any* storage type, we probably haven't seen it with S3 because they have a max file length of 1024 chars. But I've implemented a generic way to cut the length of file names. I'm not sure if it's the best way to do it.

Also for discussion: there's a line of code here when path_generator = 'flat' that sets the filename, but then that's never used (since it'll be overridden a few lines later if file_generator is random/static). 
Do we ever have a situation where we want the filename to be `slugify(path)` ? Right now the only options are `random` or `static` (hash).

@erinhmclark  - any recollection of this code change [here](https://github.com/bellingcat/auto-archiver/commit/50f4ebcdc3ab582190c2dc3c4bda750342304b69)?

### DIFF
--- a/src/auto_archiver/core/storage.py
+++ b/src/auto_archiver/core/storage.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 from abc import abstractmethod
 from typing import IO
 import os
+import platform
 
 from loguru import logger
 from slugify import slugify
@@ -43,17 +44,30 @@ class Storage(BaseModule):
     def uploadf(self, file: IO[bytes], key: str, **kwargs: dict) -> bool:
         """
         Uploads (or saves) a file to the storage service/location.
+
+        This method should not be called directly, but instead through the 'store' method,
+        which sets up the media for storage.
         """
         pass
 
     def upload(self, media: Media, **kwargs) -> bool:
+        """
+        Uploads a media object to the storage service.
+
+        This method should not be called directly, but instead be called through the 'store' method,
+        which sets up the media for storage.
+        """
         logger.debug(f'[{self.__class__.__name__}] storing file {media.filename} with key {media.key}')
         with open(media.filename, 'rb') as f:
             return self.uploadf(f, media, **kwargs)
 
     def set_key(self, media: Media, url: str, metadata: Metadata) -> None:
         """takes the media and optionally item info and generates a key"""
-        if media.key is not None and len(media.key) > 0: return
+        
+        if media.key is not None and len(media.key) > 0:
+            # media key is already set
+            return
+
         folder = metadata.get_context('folder', '')
         filename, ext = os.path.splitext(media.filename)
 
@@ -61,10 +75,8 @@ class Storage(BaseModule):
         path_generator = self.path_generator
         if path_generator == "flat":
             path = ""
-            # TODO: this is never used
-            filename = slugify(filename)  # Ensure filename is slugified
         elif path_generator == "url":
-            path = slugify(url)
+            path = slugify(url)[:70]
         elif path_generator == "random":
             path = random_str(24)
         else:
@@ -81,25 +93,7 @@ class Storage(BaseModule):
             filename = hd[:24]
         else:
             raise ValueError(f"Invalid filename_generator: {filename_generator}")
-
+        
         key = os.path.join(folder, path, f"{filename}{ext}")
-        if len(key) > self.max_file_length():
-            # truncate the path
-            max_path_length = self.max_file_length() - len(filename) - len(ext) - len(folder) - 1
-            path = path[:max_path_length]
-            logger.warning(f'Filename too long ({len(key)} characters), truncating to {self.max_file_length()} characters')
-            key = os.path.join(folder, path, f"{filename}{ext}")
 
-
-        media.key = key
-
-
-    def max_file_length(self) -> int:
-        """
-        Returns the maximum length of a file name that can be stored in the storage service.
-
-        Files are truncated if they exceed this length.
-        Override this method in subclasses if the storage service has a different maximum file length.
-        """
-        return 255 # safe max file length for most filesystems (macOS, Windows, Linux)
-
+        media._key = key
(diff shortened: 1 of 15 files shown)


src/auto_archiver/modules/local_storage/local_storage.py
```diff
@@ -20,11 +22,31 @@ def get_cdn_url(self, media: Media) -> str:
...
+
+        if len(dest) > self.max_file_length():
+            old_dest_length = len(dest)
+            filename, ext = os.path.splitext(media.key)
+            dir, filename = os.path.split(filename)
+            # see whether we should truncate filename or dir
+            if len(dir) > len(filename):
+                dir = dir[:self.MAX_FILE_LENGTH - len(self.save_to) - len(ext) - len(filename) - 1]
+            else:
+                filename = filename[:self.MAX_FILE_LENGTH - len(self.save_to) - len(ext) - len(filename) - 1]
```
msramalho, line 34: Is it the case that the too long error occurs for the totality of size `dir1/dir2/dir3/filename.txt` versus individual parts of the path `dir1TOOLONG/dir2OK/dir2OK/filenameTOOLONG.txt`? this code seems to be addressing the 1st case.

src/auto_archiver/core/storage.py
```diff
@@ -50,34 +51,55 @@ def upload(self, media: Media, **kwargs) -> bool:
...
             raise ValueError(f"Invalid filename_generator: {filename_generator}")
 
-        media.key = os.path.join(folder, path, f"{filename}{ext}")
+        key = os.path.join(folder, path, f"{filename}{ext}")
+        if len(key) > self.max_file_length():
+            # truncate the path
+            max_path_length = self.max_file_length() - len(filename) - len(ext) - len(folder) - 1
+            path = path[:max_path_length]
+            logger.warning(f'Filename too long ({len(key)} characters), truncating to {self.max_file_length()} characters')
+            key = os.path.join(folder, path, f"{filename}{ext}")
```
msramalho, line 91: this seems like a reasonable way to truncate lengthy paths

src/auto_archiver/core/storage.py
```diff
@@ -50,34 +51,55 @@ def upload(self, media: Media, **kwargs) -> bool:
...
         if media.key is not None and len(media.key) > 0: return
         folder = metadata.get_context('folder', '')
         filename, ext = os.path.splitext(media.filename)
 
         # Handle path_generator logic
-        path_generator = self.config.get("path_generator", "url")
+        path_generator = self.path_generator
         if path_generator == "flat":
             path = ""
+            # TODO: this is never used
```
erinhmclark, line 64: That can be removed here as it isn't doing anything

erinhmclark: I think it was just left in before the exception was caught for Invalid filename_generator, allowing a 'None' filename_generator option I guess?

src/auto_archiver/core/storage.py
```diff
@@ -50,34 +51,55 @@ def upload(self, media: Media, **kwargs) -> bool:
...
+
+        media.key = key
+
+
+    def max_file_length(self) -> int:
+        """
+        Returns the maximum length of a file name that can be stored in the storage service.
+
+        Files are truncated if they exceed this length.
+        Override this method in subclasses if the storage service has a different maximum file length.
```
erinhmclark, line 102: I think if we can ensure a unique filename with a max length, we can keep it consistent across storages?

erinhmclark: Or is this just in case a storage has a lower maximum file length?

erinhmclark: I'm not sure how important it is to keep the same filename across storages, but pre-modules I think these were defined in the base storage, which would have kept the generation methods consistent across them.

src/auto_archiver/modules/local_storage/local_storage.py
pjrobertson: So it's a bit complex, but windows basically only supports <260 chars. Mac and Linux support each path 'component' being up to 255 chars, with Mac having no requirement on max total length being, and linux has a MAX_PATH kernel setting which is normally 4096.

In short:
* Windows: 260-character total path limit
* macOS: No enforced total path limit, but 255 char filenames
* Linux: 4096-character total path limit, with 255 char filenames

src/auto_archiver/core/storage.py
```diff
@@ -1,11 +1,29 @@
...
+variables can be set to one of the following values:
+- 'flat': A flat structure with no subfolders
+- 'url': A structure based on the URL of the media object
+- 'random': A random structure
+
+The 'filename_generator' variable can be set to one of the following values:
+- 'random': A random string
+- 'static': A replicable strategy such as a hash
+
+If you don't want to use this naming convention, you can override the `set_key` method in your subclass.
```
erinhmclark, line 18: Do we want the generation methods to be consistent across storages? Maybe even just as a suggestion rather than an enforcement.
It would be good to allow 'core/base configs' which are set in one place, then optionally overridden, but I think this is the only place it'd be used so probably not worth it just for this!

msramalho (approved):

erinhmclark (approved):
Looks good!

src/auto_archiver/core/storage.py
msramalho: let's enforce it as is now for the ones we have/accept as PRs. this can be revisited later if something disruptive comes along.

PR #244, opened 2025-03-11
Linting and Formatting with Ruff

erinhmclark:
Add Ruff for linting and formatting.

- Add pre-commit hook for formatting (not lint fixing, this should be done manually)
- Move Makefile to project root and add commands for linting, test, docs, Docker
- Add GH Action for ruff.
- Make suggested changes from current ruff configs


pyproject.toml
```diff
@@ -57,6 +57,7 @@ dependencies = [
     "certvalidator (>=0.0.0)",
     "rich-argparse (>=1.6.0,<2.0.0)",
     "ruamel-yaml (>=0.18.10,<0.19.0)",
+    "ruff (>=0.9.10,<0.10.0)",
```
pjrobertson, line 60: Should ruff go in the 'dev' dependency?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,7 @@
+repos:
```
pjrobertson, line 1: Is this for https://pypi.org/project/pre-commit/ ?

I would vote against having another dependency (I don't see it in poetry.lock though) so will it be an option for the user to install?

pyproject.toml
erinhmclark: It should indeed!

erinhmclark:
> LGTM! Nice all round, code is definitely more readable now :)
> 
> 3 small points:
> 
> 1. I see that 2 x CI 'ruff formatting & linting' actions are being run, do we need both?
> 2. Does having a  'make docker'  make sense (to build the Dockerfile)?
> 3. Does the pre-commit.yaml need `pre-commit install` to run? Should it go in the docs?

Absolutely, that makes more sense, I have merged the lint and format steps together in the workflow.
At some point we might want to run the format from the workflow (but this requires push permissions from GH Action so I have left it for now)

I left out the build command for docker as I was using the public bellingcat image for the run, but you're right, the make commands are aimed at development it makes more sense for them to be for that - I have updated them to point at the compose file instead!

And I have added the note to the style guide on setting up pre-commit too.

pjrobertson:
Some merge conflicts here now. Also - the 2 x  CI Actions are still running. I think what you want is to change the `on` to:

```
on:
  push:
    branches: [ main ]
  pull_request:
    types: [ opened ]
```

erinhmclark:
> Some merge conflicts here now. Also - the 2 x CI Actions are still running. I think what you want is to change the `on` to:
> 
> ```
> on:
>   push:
>     branches: [ main ]
>   pull_request:
>     types: [ opened ]
> ```


With the ruff action I don't think you can join them, but I have updated it to just install ruff and run that way.
Would we only want it when opening a PR? I updated to only run on PR into main.

pjrobertson:
Yeah, I think that way of running ruff (setup python and install then run) is how I've seen most projects do it. Then we can also be more flexible about what commands we run. I only put the 'opened' in there because I wasn't sure if you had to put something inside the dict. I wonder if you can just leave that completely blank though - like this:

```
on:
  push:
    branches: [ main ]
  pull_request:
```

The rest all looks good!

pjrobertson:
Approved - but if you want to tweak the actions to run on all PRs, go ahead!

pjrobertson (approved):

PR #246, opened 2025-03-11
Better checking of cookies to add to webdriver + generic extractor tweaks

pjrobertson:
Small changes - previously I was doing a silly match for domain in cookie.domain, so 'x.com' would match 'dropbox.com'.

This is a more robust regex match.

Some other small changes:
1. Shorten the time between waits. I switched back to the 'time.sleep' method since we actually only want to wait *once* for the page to load, and then check, we don't want to wait on each item in the loop (it would cause the screenshot enricher to hang for ~25s otherwise)
2. Fix an issue I saw in the generic extractor parsing the '--cookies-from-browser' option, and made a more robust check on the ytdlp args
3. Added 'ytdlp_args' to pass additional args to ytdlp (this is separate from Erin's 'extractor_args', but it sets up the parsing to be more robust


src/auto_archiver/modules/generic_extractor/generic_extractor.py
```diff
@@ -327,19 +333,23 @@ def download(self, item: Metadata) -> Metadata:
...
             elif 'cookies_file' in auth:
                 logger.debug(f'Using cookies from file {auth["cookies_file"]} for {url}')
-                ydl_options['cookiefile'] = auth['cookies_file']
+                ydl_options.extend(('--cookies', auth['cookies_file']))
+
+        if self.ytdlp_args:
+            ydl_options += self.ytdlp_args.split(" ")
 
-        ydl = yt_dlp.YoutubeDL(ydl_options) # allsubtitles and subtitleslangs not working as expected, so default lang is always "en"
+        _, _, _, validated_options = yt_dlp.parse_options(ydl_options)
```
erinhmclark, line 351: Could use `*_, validated_options =` here for a more flexy option.

src/auto_archiver/utils/webdriver.py
```diff
@@ -48,8 +49,9 @@ def get(self, url: str):
                         self.driver.add_cookie({'name': name, 'value': value})
             elif self.cookiejar:
                 domain = urlparse(url).netloc.lstrip("www.")
```
erinhmclark, line 51: Should this be updated to removeprefix("www.") ?

src/auto_archiver/modules/generic_extractor/generic_extractor.py
```diff
@@ -327,19 +333,23 @@ def download(self, item: Metadata) -> Metadata:
...
                 logger.debug(f'Using extracted cookies from browser {auth["cookies_from_browser"]} for {url}')
-                ydl_options['cookiesfrombrowser'] = auth['cookies_from_browser']
+                ydl_options.extend(('--cookies-from-browser', auth['cookies_from_browser']))
             elif 'cookies_file' in auth:
                 logger.debug(f'Using cookies from file {auth["cookies_file"]} for {url}')
-                ydl_options['cookiefile'] = auth['cookies_file']
+                ydl_options.extend(('--cookies', auth['cookies_file']))
+
+        if self.ytdlp_args:
+            ydl_options += self.ytdlp_args.split(" ")
```
erinhmclark, line 349: Maybe they appear elsewhere or through yt-dlp but could pop a debug log here in case people accidentally overwrite/ duplicate args here.

erinhmclark (approved):
I added a few comments, but looks good!

pjrobertson:
Thanks Erin! Changes made as per your suggestions :)

erinhmclark (dismissed):
Nice! I tested out the cookies_from_browser and it's working nicely.
There's a failing download test, but that should be fixed in main.

pjrobertson:
Rebased against main, needs another approve! :)

erinhmclark (approved):

PR #247, opened 2025-03-12
Opentimestamps Module

### LINKED ISSUE #186: Test and implement new timestamping alternative with https://opentimestamps.org/
Having redundancy to the timestamp authorities implementations in the timestamping_enricher is a good thing. 
This issue is to explore https://opentimestamps.org/ and understand how reliable it can be as a new timestamping enricher by using the free "calendars"/servers they have, the code implementation seems really straightforward as they have a dedicated python library.

pjrobertson:
Closes #186 

1. Uses `python-opentimestamps` to submit to one of the calendars provided by opentimestamps
2. Checks the response to see if the timestamp is added to the calendar or pending (in most cases, it will always be pending - you can see the  Alice calendar [here](https://alice.btc.calendar.opentimestamps.org/) currently has about 7000 'pending' commitments)
3. Stores the ots file to disk, and connected to the original media
4. You can verify the OTS afterwards with `ots verify FILE.ots -f ORIGINAL_FILE`

Screen Region 2025-03-12 at 11 53 30


src/auto_archiver/modules/opentimestamps_enricher/opentimestamps_enricher.py
```diff
@@ -0,0 +1,184 @@
...
+            try:
+                # Get the file path from the media
+                file_path = media.filename
+                if not os.path.exists(file_path):
+                    logger.warning(f"File not found: {file_path}")
+                    continue
+                
+                # Create timestamp for the file - hash is SHA256
+                # Note: ONLY SHA256 is used/supported here. Opentimestamps supports other hashes, but not SHA3-512
+                # see opentimestamps.core.op
```
petertodd, line 38: FYI, for the purpose of time-stamping it'd be truly remarkable if SHA256 was ever not good enough. Even SHA1 is still secure for use as a hash function when timestamping: https://petertodd.org/2017/sha1-and-opentimestamps-proofs

src/auto_archiver/modules/opentimestamps_enricher/opentimestamps_enricher.py
```diff
@@ -0,0 +1,184 @@
...
+                            logger.debug(f"Successfully submitted to calendar: {calendar.url}")
+                            submitted_to_calendar = True
+                        except Exception as e:
+                            logger.warning(f"Failed to submit to calendar {calendar.url}: {e}")
+                    
+                    # If all calendar submissions failed, add pending attestations
+                    if not submitted_to_calendar and not timestamp.attestations:
+                        logger.info("All calendar submissions failed, creating pending attestations")
+                        for url in calendar_urls:
+                            pending = PendingAttestation(url)
```
petertodd, line 85: I may be misunderstanding what you are trying to do here. But I think this is incorrect and could cause a lot of confusion. `PendingAttestations` are for when a calendar has returned a commitment to make a timestamp in the future; they're not meant to be used as placeholders for a timestamp that failed due to unreachable calendars.

pjrobertson: You're right, we shouldn't be doing this. I'll refactor. Thanks again for the code review!

petertodd: Glad to help!

I'll let you know if the people running that IPv4-only calendar add IPv6.

src/auto_archiver/modules/opentimestamps_enricher/__manifest__.py
```diff
@@ -30,34 +25,76 @@
...
-    and secure way to prove that data existed at a certain point in time.
+    and secure way to prove that data existed at a certain point in time. A SHA256 hash of the file to be timestamped is used as the token
+    and sent to each of the 'timestamp calendars' for inclusion in the blockchain. The proof is then saved alongside the original file in a file with
+    the '.ots' extension.
 
     ### Features
-    - Creates cryptographic timestamp proofs that link files to the Bitcoin blockchain
-    - Verifies existing timestamp proofs to confirm the time a file existed
-    - Uses multiple calendar servers to ensure reliability and redundancy
+    - Creates cryptographic timestamp proofs that link files to the Bitcoin or Litecoin blockchain
```
petertodd, line 38: I'd suggest not mentioning or supporting Litecoin attestations. While the OpenTimestamps protocol technically supports them, AFAIK no-one runs calendars that timestamp via Litecoin. Nor am I aware of any code available to even do that.

It's less confusing to just leave that off entirely given it's almost certain no-one is going to take advantage of it.

src/auto_archiver/modules/opentimestamps_enricher/opentimestamps_enricher.py
```diff
@@ -0,0 +1,174 @@
...
+                
+                # Create media for the timestamp file
+                timestamp_media = Media(filename=timestamp_path)
+                # explicitly set the mimetype, normally .ots files are 'application/vnd.oasis.opendocument.spreadsheet-template'
+                timestamp_media.mimetype = "application/vnd.opentimestamps"
+                timestamp_media.set("opentimestamps_version", opentimestamps.__version__)
+                
+                verification_info = self.verify_timestamp(detached_timestamp)
+                for key, value in verification_info.items():
+                    timestamp_media.set(key, value)
```
erinhmclark, line 107: Would this store failures from the verification too?

pjrobertson: I'm not following, this will store either 'status: pending' or 'status: confirmed' if the timestamp was submitted. If no timestamp was submitted successfully, then this won't be set.

pjrobertson:
Great comments all round, thanks everyone. I've pushed some more changes.

src/auto_archiver/modules/opentimestamps_enricher/opentimestamps_enricher.py
erinhmclark: Ah okay, is it not possible for there to be no attestations at that point? aka if it's submitted there's always attestations, then these would be caught by `if not submitted_to_calendar and not timestamp.attestations:`

Does it need the conditional in that case?
```   
        if attestations:
            ...
        else:
            result["verified"] = False
```

(maybe I am misunderstanding the process, please ignore me if so!)

petertodd: > Ah okay, is it not possible for there to be no attestations at that point?

FWIW the `.ots` file format requires timestamp proofs to have at least one attestation; it's syntactically invalid if they don't, and the python-opentimestamps libraries will give you an exception. Specifically, it'll be an IOError as from the parser's point of view, the data is truncated.

pjrobertson: Yes, correct. attestations means at least 1 successful submit was made.

If there are no attestations, then there's nothing to verify, and 'opentimestamped' on the parent Meta would explicitly be set as 'False'.(see line 83)

erinhmclark: Would this be redundant code, maybe we should remove it as it would hide unexpected errors?

        else:
            result["verified"] = False

erinhmclark (approved):

PR #249, opened 2025-03-13
Move tikwm extractor into a droping for the generic extractor

pjrobertson:
Just moves over the code and refactors slightly. Also improves the logging a little bit in the generic_extractor


msramalho (approved):
tested and works as expected!

Issue #250, opened 2025-03-14
Pre-commit hook is not picking up unused f-string formatting


pjrobertson:
1. Change the code to add an unused f-string like `f"something"`
2. Try and commit
3. Note that the pre-commit hook passes successfully
4. The CI command fails 

Example:

![Image](https://github.com/user-attachments/assets/0f4cbd30-4eb7-471c-ab73-58c4e4179522)

pjrobertson:
Closed by #251

PR #251, opened 2025-03-14
Fix pre-commit for ruff check

pjrobertson:
See #250 

Also: version bump for next release


erinhmclark (approved):
The idea is though to run the ruff check (lint) commands manually as some of them require manual intervention when there are multiple refactor options - so it might be a little bit annoying in a pre-commit, but we can see how it goes!

erinhmclark (approved):

PR #254, opened 2025-03-17
Add info on building RTD versions + automated building of tagged versions

### LINKED ISSUE #252: RTD isn't saving 'different' versions in the docs
I think we need to set up automation: https://docs.readthedocs.com/platform/stable/guides/automation-rules.html

pjrobertson:
Fixes #252


erinhmclark (approved):
Nice!

PR #257, opened 2025-03-18
Standardise parse dates to get_datetime_from_str

### LINKED ISSUE #218: standardize way to parse dates
Several different approaches are currently used, between `datetime.strptime`, `dateutil.parser.parse`, 

`dateutil.parser.parse` seems to be quite robust for different formats, but perhaps a function that takes either a string or a datetime object and tries several alternatives to return a valide datetime in UTC, demo example below:

Image

some of that already lives here: https://github.com/bellingcat/auto-archiver/blob/3a93e9a5b98f58928b97f35303ea265bad118526/src/auto_archiver/core/metadata.py#L115-L126 but is not widespread and consistent.

pjrobertson:
Fixes #218


erinhmclark (approved):
Approved!

PR #259, opened 2025-03-18
Small fix for generic_extractor.py for general/ youtube extraction.

erinhmclark:
Youtube videos were getting 'no archiver' result.


pjrobertson:
Looks good! That `get_content()` vs `get('content')` (one throws an error, the other doesn't) has caught me out before. I'm not sure if there's a better way to refactor?

Approved!

pjrobertson (approved):

erinhmclark:
> Looks good! That `get_content()` vs `get('content')` (one throws an error, the other doesn't) has caught me out before. I'm not sure if there's a better way to refactor?
> 
> Approved!

So I'm not sure if we wanted to add a getter method for 'content' to Metadata, but went for just getting these fixed for now!

Issue #230, opened 2025-03-03
Validate Wacz Extractor-Enricher execution from the docker command line.


erinhmclark:
Validate the functionality of the WACZ extractor enricher and add Docker CLI execution instructions, including required environment variables, for use outside the Compose file.

pjrobertson:
this is partly due to the __manifest__ having `docker` as a `bin` requirement - which isn't the case for the docker image.

To reproduce:

1. Enable wacz_extractor_enricher in orchestration
2. Run the docker image e.g. with `docker run -it --rm -v $PWD/secrets:/app/secrets -v $PWD/local_archive:/app/local_archive bellingcat/auto-archiver https://truthsocial.com/@CNN_US/posts/110471542516802287`
3. See the error: `Module 'wacz_extractor_enricher' requires external dependency 'docker' which is not available/setup.                                  Have you installed the required dependencies for the 'wacz_extractor_enricher' module? See the README for more information.`

Issue #213, opened 2025-02-19
improve wacz file creation to handle parallel work


msramalho:
It seems that because more than one concurrent archivers were using the wacz enricher one of them failed/had issues copying over the file properly and this led to a corrupted file.

```log
worker-1  | 2025-02-19 18:15:05.332 | DEBUG    | auto_archiver.modules.wacz_enricher.wacz_enricher:enrich:87 - copying secrets/profile.tar.gz to /crawls/profile.tar.gz
worker-1  | 
worker-1  | gzip: stdin: unexpected end of file
worker-1  | tar: Unexpected EOF in archive
worker-1  | tar: Child returned status 1
worker-1  | tar: Error is not recoverable: exiting now
```

solution: have a dedicated filename for each copied over profile file

PR #261, opened 2025-03-19
Wacz minor adjustments

pjrobertson:
Should fix both #213 and #230

### DIFF
--- a/src/auto_archiver/core/module.py
+++ b/src/auto_archiver/core/module.py
@@ -237,8 +237,13 @@ class LazyBaseModule:
 
             return find_spec(dep)
 
+        def check_bin_dep(dep):
+            if dep == "docker" and os.environ.get("RUNNING_IN_DOCKER"):
+                return True
+            return shutil.which(dep)
+
         check_deps(self.dependencies.get("python", []), check_python_dep)
-        check_deps(self.dependencies.get("bin", []), lambda dep: shutil.which(dep))
+        check_deps(self.dependencies.get("bin", []), check_bin_dep)
 
         logger.debug(f"Loading module '{self.display_name}'...")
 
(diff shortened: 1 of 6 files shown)


src/auto_archiver/core/module.py
```diff
@@ -237,8 +237,13 @@ def check_python_dep(dep):
 
             return find_spec(dep)
 
+        def check_bin_dep(dep):
+            if dep == "docker" and os.environ.get("RUNNING_IN_DOCKER"):
```
erinhmclark, line 241: With wacz I think it doesn't require docker as the base image is browsertrix so it's not fully 'docker-in-docker' as it's base image is already there.  So I think if something else requires docker it wouldn't work (I'm thinking about the PO Tokens!)
Do we want to allow docker in docker in general?

msramalho: my 2cents:  let's not encourage the use of docker broadly, but I do feel this check makes sense if we want to keep the bin dependency altogether. 
We actually do docker-out-of-docker on the API side because we use a volume to connect the docker sock and so `docker` starts referring to the main OS's install and docker commands will run containers side by side vs inside one another. 

more context https://tdongsi.github.io/blog/2017/04/23/docker-out-of-docker/

pjrobertson: Yeah, the alternative is to remove the 'docker' bin requirement, which would mean a potential error for people running outside of a docker when the wacz enricher tries to spin up the docker image but docker isn't installed. 

That was my original thinking, but I thought keeping it in does actually make more sense/help, since that'd mean failing quite late in the extraction process.

pjrobertson:
Pushed a few more changes to add notes to the docs

erinhmclark (approved):
This looks good! It handles the docker/ not docker/ docker-in-docker cases nicely.

If I can get a sensible solution up with the PO Tokens I will add a developer how to guide for handling docker dependencies too.

PR #262, opened 2025-03-20
Add flexible extractor_args to generic_extractor.py

erinhmclark:
Add option to add any [extractor args](https://github.com/yt-dlp/yt-dlp/blob/master/README.md#extractor-arguments) to the generic extractor.
This was added in preparation for passing proof of origin tokens, but separating it into it's own PR.

### DIFF
--- a/tests/test_modules.py
+++ b/tests/test_modules.py
@@ -81,8 +81,20 @@ def test_load_modules(module_name):
     # check that default settings are applied
     default_config = module.configs
     assert loaded_module.name in loaded_module.config.keys()
+    defaults = {k for k in default_config}
+    assert defaults in [loaded_module.config[module_name].keys()]
+
+
+@pytest.mark.parametrize("module_name", ["local_storage", "generic_extractor", "html_formatter", "csv_db"])
+def test_config_defaults(module_name):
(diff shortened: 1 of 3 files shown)


tests/test_modules.py
```diff
@@ -82,7 +82,7 @@ def test_load_modules(module_name):
     default_config = module.configs
     assert loaded_module.name in loaded_module.config.keys()
     defaults = {k: v.get("default") for k, v in default_config.items()}
-    assert loaded_module.config[module_name] == defaults
+    assert defaults.keys() in [loaded_module.config[module_name].keys()]
```
pjrobertson, line 85: Only thing I'm wary of is that we're not actually checking the *values* of the keys now, this feels like a test regression IMO. Not sure if/when this could cause an issue, I might be over thinking.

erinhmclark: I have separated these tests out and added a little note. There are some cases (e.g. with the session_file of instagram_tbot_extractor which are meant to modify the default values on setup, so I think key to key comparison makes sense for the general case, but I have added a value test separately too.

pjrobertson (approved):
Approved!

PR #264, opened 2025-03-21
Minor fixes

pjrobertson:
This:

* Adds unit tests for utils/url
* Fixes a bug in the webdriver - it was looking for `cookies` in the auth dict, but it should be `cookie` and adds the 'removeprefix' back in, I'd accidentally removed it on the cookies check.
* Provides better logs when auth isn't supported (if a user tries to do username/password login with the screenshot enricher)

### DIFF
--- a/src/auto_archiver/utils/url.py
+++ b/src/auto_archiver/utils/url.py
@@ -4,8 +4,8 @@ from ipaddress import ip_address
 
 
 AUTHWALL_URLS = [
-    re.compile(r"https:\/\/t\.me(\/c)\/(.+)\/(\d+)"),  # telegram private channels
-    re.compile(r"https:\/\/www\.instagram\.com"),  # instagram
+    re.compile(r"https?:\/\/t\.me(\/c)\/(.+)\/(\d+)"),  # telegram private channels
+    re.compile(r"https?:\/\/(www\.)?instagram\.com"),  # instagram
 ]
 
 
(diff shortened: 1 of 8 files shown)


src/auto_archiver/utils/url.py
```diff
@@ -81,56 +81,43 @@ def is_relevant_url(url: str) -> bool:
...
+    IRRELEVANT_ENDS_WITH = [
+        ".svg",  # ignore SVGs
+        ".ico",  # ignore icons
+    ]
+
+    for end in IRRELEVANT_ENDS_WITH:
+        if clean_url.endswith(end):
+            return False
+
+    for parts in IRRELEVANT_URLS:
```
erinhmclark, line 118: Nice, this is much cleaner.

tests/utils/test_urls.py
```diff
@@ -0,0 +1,143 @@
...
+        ("https://example.com?utm_source=example", "https://example.com"),
+        ("https://example.com?utm_source=example&other=1", "https://example.com"),
+        ("https://example.com/something", "https://example.com/something"),
+        ("https://example.com/something?utm_source=example", "https://example.com/something"),
+    ],
+)
+def test_remove_get_parameters(url, without_get):
+    assert remove_get_parameters(url) == without_get
+
+    # IRRELEVANT_URLS = [
```
erinhmclark, line 73: Will you clean these up leaving them for reference?

erinhmclark (approved):
Looks good!

tests/utils/test_urls.py
pjrobertson: Yes, good catch! Done :)

erinhmclark (approved):

PR #268, opened 2025-03-24
Minor improvements

pjrobertson:


src/auto_archiver/modules/generic_extractor/generic_extractor.py
```diff
@@ -37,12 +38,24 @@ def setup(self):
...
 
             next_update_check = datetime.datetime.now() + datetime.timedelta(days=self.ytdlp_update_interval)
             with open(path, "w") as f:
                 f.write(next_update_check.isoformat())
 
+            if not updated:
+                return
+
+            if os.environ.get("AUTO_ARCHIVER_ALLOW_RESTART", "1") != "1":
+                logger.warning(
```
erinhmclark, line 51: Nice! Do we need any docs on setting the env var, or is it intended for advanced/ niche uses?

erinhmclark (approved):

src/auto_archiver/modules/generic_extractor/generic_extractor.py
pjrobertson: I thin kit's a bit too advanced for most. My thinking was that this was only really for devs!

PR #276, opened 2025-03-26
Only run 'download' actions once core completes + re-add `ubuntu-latest` to matrix

pjrobertson:
Now that the timestamping enricher is merged, we can now re-enable ubuntu-latest in the testing matrix.

Also, to cut back on the `download` tests, I've set it to only run once the core tests succeed.


pjrobertson:
P.S. I pushed one small code change here just to trigger the tests.

.github/workflows/tests-download.yaml
```diff
@@ -1,15 +1,23 @@
...
+    types:
+      - completed
+    branches: [main]
     paths:
       - src/**
 
 jobs:
   tests:
+    # Only run if the triggering workflow (tests-core) succeeded
+    if: ${{ github.event_name == 'schedule' || github.event.workflow_run.conclusion == 'success' }}
```
erinhmclark, line 20: Good plan!

src/auto_archiver/core/module.py
```diff
@@ -214,11 +214,8 @@ def load(self, config) -> BaseModule:
 
         # check external dependencies are installed
         def check_deps(deps, check):
-            for dep in deps:
-                if not len(dep):
-                    # clear out any empty strings that a user may have erroneously added
-                    continue
-                if not check(dep):
+            for dep in filter(lambda d: len(d.strip()), deps):
```
erinhmclark, line 217: good catch!

erinhmclark (approved):

PR #278, opened 2025-03-26
This force-pins cryptography to >44.0.1 to fix dependabot warning

### LINKED ISSUE #277: Review Dependabot alerts
https://github.com/bellingcat/auto-archiver/security/dependabot

Note: `cryptography` is being held back because of tsp_client, which is only compatible with `cryptography<44.0`. More info and suggested alternatives here: https://github.com/bellingcat/auto-archiver/pull/169

pjrobertson:
pyOpenSSL also no longer needed - that was needed for the old timestamping enricher and is what was keeping cryptography back.

Fixes #277 

For more info on dependabot warning, see: https://github.com/bellingcat/auto-archiver/security/dependabot/73

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -598,52 +598,60 @@ markers = {main = "sys_platform == \"win32\" or platform_system == \"Windows\"",
 
 [[package]]
 name = "cryptography"
-version = "43.0.3"
+version = "44.0.2"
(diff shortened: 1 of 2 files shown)


erinhmclark (approved):
Looks good!

PR #280, opened 2025-03-27
Download tests

pjrobertson:
I realised my previous CI changes didn't work correctly because it was set to `branches: [main]` only.

This hopefully fixes that, meaning the download CI tests run after the core tests have passed.


erinhmclark (approved):
Looks good!

pjrobertson:
Thanks! Hopefully this fixes the download tests

PR #281, opened 2025-03-28
Add InstagrAPI script

erinhmclark:
Add scripts for running a local fastAPI server, or docker instance of InstagrAPI which can be used for the Instagram API Extractor.
Previously it was expected to use the paid service https://api.instagrapi.com/docs.
These scrips use the same naming of endpoints, but only defining the paths currently used by the Auto Archiver Instagram API Extractor.

The setup process required currently involves a few steps (run server locally once to create settings file, this requires installing the requirements etc) as the first time authentication was being rejected when running in Docker.
There might be additional settings we could pass to the script which would allow verification in Docker which would simplify the process and allow the container to be setup an run with a single script, (tried with adding device and user agent, but still rejected.)


docs/source/how_to/run_instagrapi_server.md
```diff
@@ -0,0 +1,139 @@
...
+   ```
+
+6. **Watch for the message**:
+   ```
+   Login successful, session saved.
+   ```
+
+The server should now be running within that session, and accessible at  http://127.0.0.1:8000 
+
+You can set this in the orchestration.yaml file like this:
```
pjrobertson, line 64: Worth saying "in the Auto Archiver orchestration.yaml file..."?

scripts/instagrapi_server/.gitignore
pjrobertson, line 1: Is this empty file needed? Perhaps it needs the `secrets` folder, or that might already be covered by the root gitignore

pjrobertson:
Nice! Looks good to me. Only a few minor comments :)

pjrobertson:
Tests are failing here now - not sure why, could be because the poetry.lock file changed. Does that need to change?

src/auto_archiver/modules/instagram_api_extractor/__manifest__.py
```diff
@@ -31,9 +31,11 @@
         },
     },
     "description": """
-Archives various types of Instagram content using the Instagrapi API.
+Archives Instagram content using a deployment of the [Instagrapi API](https://subzeroid.github.io/instagrapi/).
 
-Requires setting up an Instagrapi API deployment and providing an access token and API endpoint.
+Requires either getting a token from using a hosted [(paid) service](https://api.instagrapi.com/docs) and setting this in the configuration file.
+Alternatively you can run your own server. We have a basic script which you can use for this which can be ran locally or using Docker.
+For more information, read the [how to guide](https://auto-archiver.readthedocs.io/en/latest/how_to/run_instagrapi_server.html) on this.
```
pjrobertson, line 38: ideally this shouldn't be hard-coded to the latest version. Should use:

`[how to guide](../../../run_instagrapi_server.md)` ( I think I got the reference right!)

pjrobertson:
```suggestion
For more information, read the [how to guide](../../../how_to/run_instagrapi_server.md) on this.
```

pjrobertson:
Just followed the guide and set everything up – it all works! Will approve once the tests are fixed :)

pjrobertson:
> > Just followed the guide and set everything up – it all works! Will approve once the tests are fixed :)
> 
> That's great to hear! I'm slightly devastated that I couldn't get it to authenticate directly in Docker, but at least this is an option 😂

I just debugged this, and I think it's because you're putting quotes around the username/password in the .env file. If I remove those, it works for me. More info: https://github.com/docker/compose/issues/3702#issuecomment-1403876840

erinhmclark:
> > > Just followed the guide and set everything up – it all works! Will approve once the tests are fixed :)
> > 
> > 
> > That's great to hear! I'm slightly devastated that I couldn't get it to authenticate directly in Docker, but at least this is an option 😂
> 
> I just debugged this, and I think it's because you're putting quotes around the username/password in the .env file. If I remove those, it works for me. More info: [docker/compose#3702 (comment)](https://github.com/docker/compose/issues/3702#issuecomment-1403876840)

Ohh that actually works! I have a script in that case, and will just update the docs to match.
That makes it so much easier to set up :)

scripts/instagrapi_server/run_instagrapi_server.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+#
+# run_instagrapi_server.sh
+# Usage:
+#   From repo root:   ./scripts/instagrapi_server/run_instagrapi_server.sh
+#   From repo root:   ./scripts/instagrapi_server/run_instagrapi_server.sh
```
pjrobertson, line 6: duplicate line

pjrobertson (approved):
Approved!

PR #304, opened 2025-05-13
Must have folder name for archive to run

djhmateer:
added into orchestration and gsheet_feeder_db as this feature is needed for my archiving.


msramalho:
Thanks for the PR, as this is an edge case and new feature, please add new test case(s) to cover its functionality to the existing tests file: https://github.com/bellingcat/auto-archiver/blob/main/tests/feeders/test_gsheet_feeder.py

PR #426, opened 2026-04-13
Add Deno to Docker image

PeterUpfold:
This adds the Deno JavaScript runtime to the container to provide a backend for `yt-dlp-ejs` to use to solve YouTube JS challenges. This enables `auto-archiver` to have access to higher quality formats than may otherwise be available.

I would welcome guidance as to whether this is the appropriate install method for this, or whether pinning to a specific release is considered more appropriate.


PeterUpfold:
Commits need tidying.

